PR #5, opened 2020-07-01
Add codecov

nsorros:


nsorros:
> You might want to add a `codecov.yml` similar to https://github.com/wellcometrust/WellcomeML/blob/master/codecov.yml, otherwise it might block you from merging if the change decreases coverage.
> 
> Other looks good to me.

Good point, i will leave it as is as a means to force me to increase coverage but if it gets annoying i will add this portion
```
coverage:
  precision: 2
  round: down
  status:
    project:
      default:
        target: auto
        threshold: 10%
```

PR #16, opened 2020-07-21
Add threshold to tag grants

nsorros:
https://trello.com/c/thI1pvAF

Tagging sample of grants for quality control by Kirstin. This PR adds the ability to tag with different thresholds, the intention is to tag with thresholds 0.3, 0.5, 0.7, 0.9 and kick off discussion about what is the appropriate default.

### DIFF
--- a/grants_tagger/tag_grants.py
+++ b/grants_tagger/tag_grants.py
@@ -55,7 +55,7 @@ if __name__ == '__main__':
         for i, grant in enumerate(yield_grants(args.grants)):
             if i % 100 == 0:
                 print(i)
-            if i & 10 != 0:
+            if i % 10 != 0:
                 # keep only 1/10th
                 continue
             tagged_grant = tag_grants(grant, args.threshold)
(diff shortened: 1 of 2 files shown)


nsorros:
[Tag grants with new ensemble model for quality check from Kirstin. Use four different threshold to kick discussion about better default](https://trello.com/c/thI1pvAF/1036-tag-grants-with-new-ensemble-model-for-quality-check-from-kirstin-use-four-different-threshold-to-kick-discussion-about-better-d)

grants_tagger/tag_grants.py
```diff
@@ -15,7 +15,7 @@ def yield_grants(grants_path):
         for grant in csv_reader:
             yield grant
 
-def tag_grants(grant):
+def tag_grants(grant, threshold):
```
lizgzil, line 18: is this going to break anything by not putting in a default value? (i.e. is tag_grants being used anywhere else without the threshold argument)

nsorros: not but good point, for backwards compatibility i should add a default

nsorros:
> Maybe you can tag also with Liz's formula (F1_Score/2)

this is if you want to maximise f1

nsorros:
Adding tests is not straightforward as explained in #19

PR #21, opened 2020-08-25
Predict mesh

nsorros:
This PR exposes a predict function for tagging grants with mesh. It also creates a function to tag grants which is used for quality controls but could be more closely align with the function / task that runs in airflow / argo.

It also updated the version to reflect these changes which will be required when integrating with argo.

In the spirit of increasing test coverage I wrote tests which make a heavy use of temp files due to the nature of the functions i am testing. If there is a better way to write those tests by refactoring some code do say. A bit of the test code is repeated as well which I think is common in tests but again if there is a better way i am keen to hear it.

### DIFF
--- a/grants_tagger/predict_mesh.py
+++ b/grants_tagger/predict_mesh.py
@@ -1,0 +1,37 @@
+"""
+"""
+import pickle
+import os
+
+from scipy.sparse import hstack
+
+
+def predict_mesh_tags(X, model_path, label_binarizer_path,
+                      probabilities=False, threshold=None):
+    # TODO: generalise tfidf to vectorizer.pkl
+    with open(f"{model_path}/tfidf.pkl", "rb") as f:
+        vectorizer = pickle.loads(f.read())
+    with open(label_binarizer_path, "rb") as f:
+        label_binarizer = pickle.loads(f.read())
+
+    Y_pred_proba = []
+    for tag_i in os.listdir(model_path):
+        if tag_i == "vectorizer.pkl":
+            continue
+        with open(f"{model_path}/{tag_i}.pkl", "rb") as f:
+            classifier = pickle.loads(f.read())
+        X_vec = vectorizer.transform(X)
+        Y_pred_i = classifier.predict_proba(X_vec)
+        Y_pred_proba.append(Y_pred_i)
+    Y_pred_proba = hstack(Y_pred_proba)
+    
+    tags = []
+    for y_pred_proba in Y_pred_proba:
+        tags_i = [tag for tag, prob in zip(label_binarizer.classes_, y_pred_proba) if prob > 0.5]
+        tags.append(tags_i)
+    return tags
+
+if __name__ == '__main__':
+    X = ["malaria"]
+    tags = predict_mesh_tags(X, "disease_mesh_tfidf-svm-2020.07.0/", "models/")
+    print(tags)
--- a/grants_tagger/predict_mesh.py
+++ b/grants_tagger/predict_mesh.py
@@ -1,4 +1,7 @@
 """
+Predict function for disease part of mesh that optionally
+exposes probabilities and that you can set the threshold 
+for making a prediction
 """
 import pickle
 import os
@@ -32,8 +35,3 @@ def predict_mesh_tags(X, model_path, label_binarizer_path,
             tags_i = [tag for tag, prob in zip(label_binarizer.classes_, y_pred_proba) if prob > threshold]
         tags.append(tags_i)
     return tags
-
-if __name__ == '__main__':
-    X = ["malaria", "ebola"]
-    tags = predict_mesh_tags(X, "models/disease_mesh_tfidf-svm-2020.07.0/", "models/disease_mesh_label_binarizer.pkl", probabilities=True)
-    print(tags)
--- a/grants_tagger/predict_mesh.py
+++ b/grants_tagger/predict_mesh.py
@@ -10,7 +10,8 @@ from numpy import hstack
 
 
 def predict_mesh_tags(X, model_path, label_binarizer_path,
-                      probabilities=False, threshold=0.5):
+                      probabilities=False, threshold=0.5,
+                      y_batch_size=512):
     # TODO: generalise tfidf to vectorizer.pkl
     with open(f"{model_path}/tfidf.pkl", "rb") as f:
         vectorizer = pickle.loads(f.read())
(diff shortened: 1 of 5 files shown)


nsorros:
[Deploy tfidf-svm baseline for disease Mesh](https://trello.com/c/b5b0rSwQ/1021-deploy-tfidf-svm-baseline-for-disease-mesh)

setup.py
```diff
@@ -21,7 +21,7 @@ def package_files(directory):
     description='A machine learning model to tag grants',
     packages=find_packages(),
     include_package_data=True,
-    version='2020.07.1',
+    version='2020.07.2',
     package_data={'': extra_files},
```
jdu, line 25: If you look here https://github.com/wellcometrust/rori/blob/master/setup.py#L23 at how we have entry_points set up on rori, it means that when you install grants_tagger onto a system, it will make an alias in /usr/bin/<name> so that you can run the package as a CLI tool, then you don't need to wrap the library in an outer python script to run it.

You can just to `<entrypoint_name> --<arg> <value>` without having to invoke python or anything.

Isn't necessary now, but worth noting for libs like this which have a CLI interface defined.

nsorros: Great point and will definitely develop that down the line.

tests/test_tag_grants_with_mesh.py
```diff
@@ -0,0 +1,77 @@
...
+            f.write(pickle.dumps(model))
+
+def test_tag_grants_with_mesh():
+    with tempfile.TemporaryDirectory() as tmp_dir:
+        model_path = f"{tmp_dir}/model/"
+        os.mkdir(model_path)
+        label_binarizer_path = f"{tmp_dir}/label_binarizer.pkl"
+        train_test_model(model_path, label_binarizer_path)
+
+        with tempfile.NamedTemporaryFile(mode="w+") as tmp_grants:
```
jdu, line 58: You could potentially ditch using NamedTemporaryFile here as you've alread created a TemporaryDirectory above , in order to simplify things. Just create your file in the already created Temporary Directory and it will get cleaned up when the encapsulating context passes out of scope (i.e. the tmp_dir variable).

Admittedly it would only save a little overhead but you would then not have to reason about temp files as anything in that tmp_dir woudl get wiped out anyway.

grants_tagger/predict_mesh.py
```diff
@@ -0,0 +1,37 @@
...
+import os
+
+from numpy import hstack
+
+
+def predict_mesh_tags(X, model_path, label_binarizer_path,
+                      probabilities=False, threshold=0.5):
+    # TODO: generalise tfidf to vectorizer.pkl
+    with open(f"{model_path}/tfidf.pkl", "rb") as f:
+        vectorizer = pickle.loads(f.read())
```
aCampello, line 16: Is this better than `pickle.load(f)` ?

tests/test_tag_grants_with_mesh.py
nsorros: Good point, this is what i did in the other test and i was wondering whether it was hacky or standard.

grants_tagger/predict_mesh.py
```diff
@@ -0,0 +1,37 @@
...
+    with open(label_binarizer_path, "rb") as f:
+        label_binarizer = pickle.loads(f.read())
+
+    nb_labels = len(label_binarizer.classes_)
+    Y_pred_proba = []
+    for tag_i in range(0, nb_labels, 512):
```
aCampello, line 22: Is 512 a default batch size? In this case you could perhaps get `y_batch_size` as an argument in line with your train function (https://github.com/wellcometrust/grants_tagger/blob/50b2f09011c3f5475a6af5fccb58dd566ddb3452/grants_tagger/train.py#L186)

setup.py
aCampello: I am interested to see that too (but got 404 on tori :( ). Is it the same as the flask setup.py?

grants_tagger/tag_grants_with_mesh.py
```diff
@@ -0,0 +1,67 @@
+"""
+Tag grants with MeSH tags
+"""
+from argparse import ArgumentParser
+from pathlib import Path
+import csv
+
+from grants_tagger.predict_mesh import predict_mesh_tags as predict_tags
+
+MODEL_PATH = "models/disease_mesh_tfidf-svm-2020.07.0/"
```
aCampello, line 10: perhaps receive those as a default value argument in ArgParser?

aCampello (approved):
Looks good, with some minor suggestions.

grants_tagger/predict_mesh.py
nsorros: Not better but i remember it better :P

grants_tagger/predict_mesh.py
nsorros: indeed, good point

grants_tagger/tag_grants_with_mesh.py
nsorros: possibly better yes

setup.py
jdu: Ah sorry, it just looks like this

```
setup(
    ...
    entry_points={
        "console_scripts": [
            "rori=rori.main:server"
        ]
    },
    ...
)
```

Where you want a CLI tool available to the user as `rori` which maps to the rori{package}.main{pythonfile}:server{method}

When this package is installed, you can start up rori by just entering the command `rori --config /etc/rori/config.yaml` on the command-line.

tests/test_tag_grants_with_mesh.py
jdu: You can do it the way you've currently done (i.e. leave as is) if you want, it's just you can trim the excess overhead of managing another temp handle and reduce the amount of cleanup required by the library at the end of it's run by using the dir. I don't think there's a whole lot of guidance on whether it's better to do it the way you have or to use re-use the dir. Just wanted to point out that you could have.

grants_tagger/tag_grants_with_mesh.py
nsorros: was just keeping in line with science tagger that hard codes the model that we use in production which is packed in the wheel. in this case we do not pack the model, it makes more sense to have it as an arg

grants_tagger/predict_mesh.py
nsorros: 8ce1ab896ba8c021dc9fd3ee7b90fb3c32a95982

grants_tagger/tag_grants_with_mesh.py
nsorros: a7fa33452ac237d6f2c7113d410967fd04046582

tests/test_tag_grants_with_mesh.py
nsorros: c1e0c692c941b7dc826aeaa287f99a59d9d3ca85

PR #25, opened 2020-09-03
Add faster predict and tests

nsorros:
Adds a new predict function for science tagging that work on multiple grants at the same time which speeds up inference because it allows for matrix multiplications to happen.

At the same time we add a similar to predict_mesh test which makes use of a test model. To do so we had to move away from hard coding the models to pass them as params. As for science tagging those models are packaged into the wheel by default the function can find them unlike mesh that models need to become available via a different route.

### DIFF
--- a/grants_tagger/predict.py
+++ b/grants_tagger/predict.py
@@ -63,6 +63,30 @@ def predict_tags(x, probabilities=False, threshold=0.5,
         tags = [tag for tag, prob in zip(tag_names, Y_pred_probs[0]) if prob > threshold]
     return tags
 
+def predict_tags_fast(X, probabilities=False, threshold=0.5):
+    label_binarizer=DEFAULT_LABELBINARIZER
+
+    Y_pred_probs = np.zeros((len(X), len(label_binarizer.classes_)))
+    for model_ in [DEFAULT_SCIBERT, DEFAULT_TFIDF_SVM]:
+        Y_pred_probs_model = model_.predict_proba(X)
+        Y_pred_probs += Y_pred_probs_model
+
+        Y_pred_probs /= 2
+
+    tag_names = label_binarizer.classes_
+
+    if probabilities:
+        tags = [
+            {tag: prob for tag, prob in zip(tag_names, Y_pred_prob)}
+            for Y_pred_prob in Y_pred_probs
+        ]
+    else:
+        tags = [
+            [tag for tag, prob in zip(tag_names, Y_pred_prob) if prob > threshold]
+            for Y_pred_prob in Y_pred_probs
+        ]
+    return tags
+
 if __name__ == "__main__":
     argparser = argparse.ArgumentParser()
     argparser.add_argument(
--- a/grants_tagger/predict.py
+++ b/grants_tagger/predict.py
@@ -19,56 +19,23 @@ DEFAULT_SCIBERT_PATH = os.path.join(FILEPATH, '../models/scibert-2020.05.5')
 DEFAULT_TFIDF_SVM_PATH = os.path.join(FILEPATH, '../models/tfidf-svm-2020.05.2.pkl')
 DEFAULT_LABELBINARIZER_PATH = os.path.join(FILEPATH, '../models/label_binarizer.pkl')
 
-DEFAULT_SCIBERT = BertClassifier()
-DEFAULT_SCIBERT.load(DEFAULT_SCIBERT_PATH)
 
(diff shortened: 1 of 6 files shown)


nsorros:
Testing predict is a bit tricky as the models are not packaged and sent to test

grants_tagger/predict.py
```diff
@@ -19,57 +19,54 @@
...
-                 label_binarizer=DEFAULT_LABELBINARIZER):
-    '''Input example text when running the .py file in the terminal to return predicted grant tags - use format:'''
-    if type(model) == list: # ensemble of models
-        Y_pred_probs = np.zeros((1, len(label_binarizer.classes_)))
-        for model_ in model:
-            Y_pred_probs_model = model_.predict_proba([x])
-            Y_pred_probs += Y_pred_probs_model
+    Y_pred_probs = np.zeros((len(X), len(label_binarizer.classes_)))
+    for model in [scibert, tfidf_svm]:
+        Y_pred_probs_model = model.predict_proba(X)
```
aCampello, line 38: Just a note to myself that this loops assumes that the labels in both models are completely aligned and mapped to the same classes, which means you used the same binarizer.

grants_tagger/predict.py
```diff
@@ -19,57 +19,54 @@
...
-    """
-    data = zip(tags.tolist()+['------ THRESHOLD ------'], probs.tolist()+[threshold])
-    sorted_tags_probs = sorted(data, key = itemgetter(1), reverse=True)
-    return sorted_tags_probs
-
-def predict_tags(x, probabilities=False, threshold=0.5,
-                 model=[DEFAULT_SCIBERT, DEFAULT_TFIDF_SVM],
-                 label_binarizer=DEFAULT_LABELBINARIZER):
-    '''Input example text when running the .py file in the terminal to return predicted grant tags - use format:'''
-    if type(model) == list: # ensemble of models
```
aCampello, line 48: Removing this `if` you are assuming that model will always be a list now, losing your ability to predict against just one model.

grants_tagger/tag_grants.py
```diff
@@ -27,17 +27,19 @@ def tag_grants(grant, threshold):
...
-        'Grant ID': grant['grant_id'],
-        'Reference': grant['reference'],
-        'Grant No.': grant['grant_no']
-    }
-    tagged_grant.update({
-        f"Science Category#{i+1}": tag
-        for i, tag in enumerate(tags)
-    })
-    return tagged_grant
+    grants_texts = [grant["title"]+["synopsis"] for g in grants]
```
aCampello, line 30: I am confused with this syntax. It looks like it won't work, because `g` wasn't defined. Should it be `for grant in grants]` ?

grants_tagger/tag_grants.py
```diff
@@ -52,11 +54,13 @@ def tag_grants(grant, threshold):
         fieldnames += [f"Disease Category#{i}" for i in range(1,6)]
         csv_writer = csv.DictWriter(f_o, fieldnames=fieldnames)
```
aCampello, line 55: DictWriter ftw!!

aCampello (changes requested):
Improvements look good (in particular predicting the whole lot!). I am either reading it wrongly, or it seems that there is a broken line of code:

https://github.com/wellcometrust/grants_tagger/blob/a9557a2f80e7ed09059f0d991b8629968a62942a/grants_tagger/tag_grants.py#L30

Can you have a look and modify?

grants_tagger/predict.py
nsorros: good point, maybe a comment to add

grants_tagger/tag_grants.py
nsorros: Niice, well i did not write a test for this one, one more reminder why tests should be added for all code you introduce. Will add now.

nsorros:
> Improvements look good (in particular predicting the whole lot!). I am either reading it wrongly, or it seems that there is a broken line of code:
> 
> https://github.com/wellcometrust/grants_tagger/blob/a9557a2f80e7ed09059f0d991b8629968a62942a/grants_tagger/tag_grants.py#L30
> 
> Can you have a look and modify?

Resolved by 81abdda2c753ec4ec53e09680db9c8d028ce9b93

grants_tagger/tag_grants.py
nsorros: 81abdda2c753ec4ec53e09680db9c8d028ce9b93

aCampello (approved):
Looks good :shipit:

PR #26, opened 2020-09-10
Neural network experiments on Mesh

nsorros:
This contains most of the code needed to run and reproduce the experiments I want to run. There are still a few changes I will need to make as I go about the experiments but most of the functionality is there.

A lot of the code is similar to other parts like the train loop but I wanted to keep some flexibility for the experiments to be able to easily try different params without having the interdependencies. Naturally trying different params and train are similar but also different processes.

Lastly this is also not so much about hyper parameter optimisation at this point and that is why I have not looked into using sklearn Gridsearch etc. Again I want the flexibility to try some theories for example that as more tags are being predicted more steps of gradient descent are needed.

### DIFF
--- a/scripts/run_mesh_experiments.py
+++ b/scripts/run_mesh_experiments.py
@@ -1,4 +1,9 @@
+"""
+Experiments on disease mesh tags to understand role of params
+"""
+from argparse import ArgumentParser
 from datetime import datetime
+from pathlib import Path
 import json
 import math
 
@@ -11,9 +16,6 @@ from tqdm import tqdm
 import tensorflow as tf
 import numpy as np
 
-DEFAULT_VOCABULARY_SIZE = 400_000
-DEFAULT_SEQUENCE_LENGTH = 400
-DEFAULT_NB_TAGS = 512
 
 def create_dataset(data_path, nb_tags, nb_examples_per_tag):
     texts = []
@@ -39,9 +41,7 @@ def create_dataset(data_path, nb_tags, nb_examples_per_tag):
                 texts.append(item["text"])
                 tags.append(item_tags)
             if all([c > nb_examples_per_tag for t, c in tags_count.items()]):
-                print(tags_count)
                 break
-        print(tags_count)
         return texts, tags
 
 def vectorize_data(train_texts, train_tags, test_texts, test_tags,
@@ -58,9 +58,9 @@ def vectorize_data(train_texts, train_tags, test_texts, test_tags,
     Y_test = label_binarizer.transform(test_tags)
     return X_train, X_test, Y_train, Y_test
 
-def build_model(learning_rate=0.01, batch_size=256, attention=True,
-                vocabulary_size=400_000, sequence_length=400, nb_tags=512,
-                l2=1e-6, dropout=0.1, hidden_size=100, dense_size=100_000):
+def build_cnn_model(learning_rate=0.01, batch_size=256, attention=True,
+                    vocabulary_size=400_000, sequence_length=400, nb_tags=512,
+                    l2=1e-6, dropout=0.1, hidden_size=100, dense_size=100_000):
     model = CNNClassifier(
         attention=attention, multilabel=True,
         learning_rate=learning_rate, batch_size=batch_size,
@@ -68,6 +68,7 @@ def build_model(learning_rate=0.01, batch_size=256, attention=True,
         dropout=dropout, l2=l2
     )
     model = model._build_model(vocab_size=vocabulary_size, sequence_length=sequence_length, nb_outputs=nb_tags)
+    print(model.summary())
     return model
 
 def build_bilstm_model(learning_rate=0.01, batch_size=256, attention=True,
@@ -140,7 +141,7 @@ def train(X_train, X_test, Y_train, Y_test, params):
     validation_steps = math.ceil(X_test.shape[0]/batch_size)
 
     if architecture == 'cnn':
-        build_model_f = build_model
+        build_model_f = build_cnn_model
     elif architecture == 'bilstm':
         build_model_f = build_bilstm_model
     else:
@@ -178,7 +179,6 @@ def experiment(data_path, params):
     sequence_length = params["sequence_length"]
     texts, tags = create_dataset(
         data_path, nb_tags, nb_examples_per_tag)
-    print(len(texts))
 
     nb_train = len(texts) - min(int(0.2*len(texts)), 10_000)
     train_texts = texts[:nb_train]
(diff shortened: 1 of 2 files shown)


scripts/run_mesh_experiments.py
```diff
@@ -0,0 +1,293 @@
...
+    # vectorize tags
+    label_binarizer = MultiLabelBinarizer(sparse_output=True)
+    label_binarizer.fit(train_tags)
+    Y_train = label_binarizer.transform(train_tags)
+    Y_test = label_binarizer.transform(test_tags)
+    return X_train, X_test, Y_train, Y_test
+
+def build_cnn_model(learning_rate=0.01, batch_size=256, attention=True,
+                    vocabulary_size=400_000, sequence_length=400, nb_tags=512,
+                    l2=1e-6, dropout=0.1, hidden_size=100, dense_size=100_000):
```
nsorros, line 63: dense_size default should be 10k :p

nsorros:
@aCampello @lizgzil any thoughts on this one?

nsorros:
> @aCampello @lizgzil any thoughts on this one?

@aCampello @lizgzil

aCampello:
Let me have a look

nsorros:
> Let me have a look

I love how carefully you are looking over it 😂

scripts/run_mesh_experiments.py
```diff
@@ -0,0 +1,293 @@
...
+    argparser.add_argument('--hidden_size', type=int, default=100)
+    argparser.add_argument('--dense_size', type=int, default=10_000)
+    argparser.add_argument('--attention', type=bool, default=True)
+    argparser.add_argument('--architecture', default="cnn")
+    argparser.add_argument('--epochs', type=int, default=500)
+    argparser.add_argument('--vocabulary_size', type=int, default=400_000)
+    argparser.add_argument('--sequence_length', type=int, default=400)
+    argparser.add_argument('--config', type=Path)
+    args = argparser.parse_args()
+
```
aCampello, line 243: So you have two ways of submitting a model - via the config file or via setting the parameters yourself. I see, that sounds good.

scripts/run_mesh_experiments.py
```diff
@@ -0,0 +1,293 @@
...
+def experiment(data_path, params):
+    nb_tags = params["nb_tags"]
+    nb_examples_per_tag = params["nb_examples_per_tag"]
+    vocabulary_size = params["vocabulary_size"]
+    sequence_length = params["sequence_length"]
+    texts, tags = create_dataset(
+        data_path, nb_tags, nb_examples_per_tag)
+
+    nb_train = len(texts) - min(int(0.2*len(texts)), 10_000)
```
aCampello, line 183: So you are saying that the test set will have a maximum of 10000 docs? Why is that?

aCampello (approved):
I apologise for the delay in reviewing! I appreciate it is experiment code so it might be repeated from some parts of the project, so it looks good to me. I have some small suggestions and comments.

scripts/run_mesh_experiments.py
nsorros: you mean the 100_000. it is python notation to make it more readable. https://www.python.org/dev/peps/pep-0515/

scripts/run_mesh_experiments.py
```diff
@@ -0,0 +1,293 @@
...
+from tensorboard.plugins.hparams import api as hp
+from sklearn.preprocessing import MultiLabelBinarizer
+from sklearn.metrics import f1_score
+from scipy.sparse import vstack, csr_matrix
+from tqdm import tqdm
+import tensorflow as tf
+import numpy as np
+
+
+def create_dataset(data_path, nb_tags, nb_examples_per_tag):
```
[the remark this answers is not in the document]
nsorros: I think this is the only new function :p

scripts/run_mesh_experiments.py
```diff
@@ -0,0 +1,293 @@
...
+    texts = []
+    tags = []
+
+    tags_count = {}
+    with open(data_path) as f:
```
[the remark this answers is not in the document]
nsorros: Yep it is pretty hard to understand, might rewrite it or add comments

scripts/run_mesh_experiments.py
```diff
@@ -0,0 +1,293 @@
...
+        dropout=dropout
+    )
+    model = model._build_model(vocab_size=vocabulary_size, sequence_length=sequence_length, nb_outputs=nb_tags)
+    print(model.summary())
+    return model
+
+def build_custom_bilstm_model(learning_rate=0.01, batch_size=256, attention=True,
+                              vocabulary_size=400_000, sequence_length=400, nb_tags=512,
+                              l2=1e-6, dropout=0.1, hidden_size=100, dense_size=100_000):
+    bias_init = 0 #np.log(Y_train.sum(axis=0) / (Y_train.shape[0] - Y_train.sum(axis=0)))
```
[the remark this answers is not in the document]
nsorros: Well, this is in theory the correct initialisation if you have imbalanced data. It should speed up learning something I have not observed yet. Check here https://www.tensorflow.org/tutorials/structured_data/imbalanced_data#optional_set_the_correct_initial_bias

scripts/run_mesh_experiments.py
nsorros: I do that in all my scripts, nothing new :)

scripts/run_mesh_experiments.py
nsorros: Well I should have done this earlier tbh. Andrew Ng suggests that after a dataset size you do not need the typical 80/20 split but more of a 99/1. You really only need enough data to make a statistically robust estimate of the error or accuracy. Check https://www.coursera.org/lecture/machine-learning-projects/size-of-the-dev-and-test-sets-HOby4

aCampello: Yeah, I remember that from the course. The one thing to bear in mind though is that you have an extreme mutilabel problem, so a statistically robust sample might require more data. (maybe 10.000 is fine)

aCampello (approved):

scripts/run_mesh_experiments.py
nsorros: It is a good point. i am actually using the same number used by deep mesh https://www.ncbi.nlm.nih.gov/pmc/articles/PMC4908368/pdf/btw294.pdf
```
We sorted all downloaded citations by time. The latest 10 000 citations were used as validation set for tuning parameters of binary classifier
```
But keen to hear if there is a way or rule of thumb to decide how many examples are enough.

aCampello:
This his approved long ago. Maybe we could merge?

nsorros:
Closing this as it is not scoped for Q1. Might reopen in the future.

PR #27, opened 2020-09-16
Add stream option for BiLSTM and CNN approach

nsorros:


nsorros:
This will be implemented on the WellcomeML side since if at a later point sparse tensors become available then there will be no change needed in the train loop. The yielding implemented is to circle the absence of proper support for sparse target labels in tensorflow.

PR #28, opened 2020-09-23
Add SciSpacy mesh tagger

nsorros:
SciSpacy released a new version that is able to tag with MeSH. Essentially, what is happening is that scispacy detects entities and then links them to mesh. This is not the same problem as tagging but it is worth comparing how well it performs.

https://trello.com/c/OIcRgVem


aCampello:
Is it ready to review?

nsorros:
> Is it ready to review?

I need to add the test that mentioned on Slack and then it will be. I added you a bit before it was ready, sorry 😬

nsorros:
@aCampello maybe when you finish with the other review you can move here :P or @lizgzil can take this one. Who would it be let's see.

lizgzil:
@nsorros I'm really busy on the reach evaluation stuff to be able to look at this before the end of today :( might be able to tomorrow morning!

nsorros:
One thing I am not sure of is where is the code coverage report? 🤔

nsorros:
@lizgzil @aCampello

Makefile
```diff
@@ -19,19 +19,33 @@ sync_artifacts:
...
+	echo "⚠️  Virtualenv is broken and needs to be reinstalled"
 
 .PHONY: test
-test:
+test: virtualenv
 	$(VIRTUALENV)/bin/python -m spacy download en_trf_bertbaseuncased_lg
-	$(VIRTUALENV)/bin/pytest --disable-warnings -v --cov=grants_tagger
+	$(VIRTUALENV)/bin/pytest --disable-warnings -v --cov=grants_tagger -m "not scispacy"
+
+.PHONY: test_scispacy
```
lizgzil, line 45: FYI when I did `make test_scispacy` I got 
```
ERROR: Command errored out with exit status 128: git clone -q git://github.com/wellcometrust/WellcomeML.git /Users/gallaghe/Code/grants_tagger/venv/src/wellcomeml Check the logs for full command output.
```
Could you just have `wellcomeml[deep-learning]` in the requirements now or did you need that specific wellcomeml version?

When I did replace this with `wellcomeml[deep-learning]` then `make test_scispacy` passed

tests/test_scispacy_meshtagger.py
```diff
@@ -0,0 +1,35 @@
...
+X = [
+    "Malaria is a mosquito-borne infectious disease that affects humans and other animals. Malaria causes symptoms that typically include fever, tiredness, vomiting, and headaches. In severe cases it can cause yellow skin, seizures, coma, or death",
+    "The human immunodeficiency viruses are two species of Lentivirus that infect humans. Without treatment, average survival time after infection with HIV is estimated to be 9 to 11 years, depending on the HIV subtype."
+]
+
+@pytest.mark.scispacy
+def test_fit():
+    from grants_tagger.scispacy_meshtagger import SciSpacyMeshTagger
+    scispacy_meshtagger = SciSpacyMeshTagger(mesh_tags = MESH_TAGS)
+    scispacy_meshtagger.fit()
```
lizgzil, line 17: Don't you need an `assert` here or is the failure of this fitting the test?

grants_tagger/scispacy_meshtagger.py
```diff
@@ -0,0 +1,59 @@
...
+    def fit(self, *_):
+        nlp = spacy.load("en_core_sci_sm")
+        abbreviation_resolver = AbbreviationDetector(nlp)
+        nlp.add_pipe(abbreviation_resolver)
+        linker = EntityLinker(resolve_abbreviations=True, name="mesh")
+        nlp.add_pipe(linker)
+        self.nlp = nlp
+
+    def _get_tags(self, doc):
+        mesh_ents = defaultdict(int)
```
lizgzil, line 33: is `mesh_ent_prob` an int then? I assumed it was a probability from the name

grants_tagger/scispacy_meshtagger.py
```diff
@@ -0,0 +1,59 @@
...
+        for ent in doc.ents:
+            if ent._.kb_ents and ent._.kb_ents[0][0] in self.mesh_tags:
+                mesh_ent = ent._.kb_ents[0]
+                mesh_ent_ui = mesh_ent[0]
+                mesh_ent_prob = mesh_ent[1]
+                mesh_ents[mesh_ent_ui] += mesh_ent_prob
```
lizgzil, line 39: I wonder whether you mean if the probability of this mesh term is >0.5 this will be adding 1 and otherwise will be adding 0 (due to `mesh_ents = defaultdict(int)`)? If this is the case I think you need a comment to clear it up

lizgzil: or even 
```
if mesh_ent_prob > 0.5:
   mesh_ents[mesh_ent_ui] += 1

File: grants_tagger/scispacy_meshtagger.py [hunk 5]
```diff
@@ -0,0 +1,59 @@
...
+from scipy.sparse import csr_matrix
+import scispacy
+import spacy
+
+
+class SciSpacyMeshTagger():
+    def __init__(self, mesh_tags, threshold=2):
+        """
+        mesh_tags: list of mesh UI descriptros e.g. [D00001]
+        """
```
lizgzil, line 20: could do with a description of what `threshold` is

scripts/evaluate_scispacy_meshtagger.py
```diff
@@ -0,0 +1,40 @@
...
+    disease_mesh_data = pd.read_csv(mesh_tags_path)
+    mesh_name2tag = {name: tag for tag, name in zip(disease_mesh_data['DescriptorUI'],
+                                                    disease_mesh_data['DescriptorName'])}
+
+    mesh_tags_names = label_binarizer.classes_
+    mesh_tags = [mesh_name2tag[name] for name in mesh_tags_names]
+
+    X, Y, _ = load_data(mesh_data_path, label_binarizer)
+    X = X[:10_000]
+    Y = Y[:10_000, :]
```
lizgzil, line 27: Sure it wont be biased to take the first ones as your subset rather than random?

lizgzil (commented):
Looks good, I think you need to do something about how wellcomeml is imported, see my comment

Makefile
nsorros: This is a good point, the version is pinned to have reproducible results. But in order to run this code you need to unpin it indeed. I think I should have just mentioned it in the description. It is intentionally as is.

grants_tagger/scispacy_meshtagger.py
nsorros: hmm, this sounds like a bug, well spotted, let me double check

grants_tagger/scispacy_meshtagger.py
nsorros: this is probably a bug, will investigate. well spotted

scripts/evaluate_scispacy_meshtagger.py
nsorros: :p you don't like taking the first 10k i sense. i might create a script that creates a sample tbh, it might be useful for comparing against MTI which is yet another tool

tests/test_scispacy_meshtagger.py
nsorros: if it can fit without throwing an exception we are good

nsorros:
[Calculate Scispacy accuracy for Mesh](https://trello.com/c/OIcRgVem/1086-calculate-scispacy-accuracy-for-mesh)

Makefile
nsorros: @lizgzil are you sure you runned `make test_scispacy`? do you mind running again as it passes locally and in the tests.

lizgzil: it looks like that error was actually when I ran `make test` 🤦‍♀️  , `make test_scispacy` did pass (so will approve), but I got the error/warning:
```
ERROR: After October 2020 you may experience errors when installing or updating packages. This is because pip will change the way that it resolves dependency conflicts.

We recommend you use --use-feature=2020-resolver to test your packages with the new resolver before it becomes the default.

wellcomeml 2020.9.0 requires spacy==2.2.1, but you'll have spacy 2.3.2 which is incompatible.
spacy-transformers 0.5.1 requires spacy<2.3.0,>=2.2.1, but you'll have spacy 2.3.2 which is incompatible.
```
so maybe you'll need to do that change soon

lizgzil (approved):
I will approve, but leave it to you to decide if my comment on `mesh_ents[mesh_ent_ui] += mesh_ent_prob` was a bug or not

Makefile
nsorros: in order to run make test you need to recreate the environment as the scispacy installation breaks it. Check https://github.com/wellcometrust/grants_tagger/blob/1e8ff1e04722f6c431368127fca2fd321bcb9afc/Makefile#L38

nsorros:
> I will approve, but leave it to you to decide if my comment on `mesh_ents[mesh_ent_ui] += mesh_ent_prob` was a bug or not

investigating and if a bug will fix

grants_tagger/scispacy_meshtagger.py
nsorros: 99d9c7e843ce7696eb62e9b7132f5d66cf7d3d37

grants_tagger/scispacy_meshtagger.py
nsorros: Turns out that python comes to the rescue and converts to float. Int is used only to initialise in case no key is found.
```
>>> from collections import defaultdict
>>> a = defaultdict(int)
>>> a['b']
0
>>> a['b'] += .08
>>> a
defaultdict(<class 'int'>, {'b': 0.08})
```

nsorros: Slighly more information here https://stackoverflow.com/questions/5900578/how-does-collections-defaultdict-work. The argument passed, in this case int, is used to initialise the elements that are not present. Then normal python takes over and when you add a float to an int, python converts the int to float which what happens here

lizgzil: ok great , good to know!

scripts/evaluate_scispacy_meshtagger.py
nsorros: 5428df6c20fb41ac8c57055e98031bddb47305aa and 21bf6e79839b27cb56093540a76ab176a23057ca

scripts/create_mesh_sample.py
```diff
@@ -0,0 +1,43 @@
...
+
+def count_lines(jsonl_path):
+    counter = 0
+    with open(jsonl_path) as f:
+        for line in f:
+            counter += 1
+    return counter
+
+def yield_sample_data(data_path, sample_size=SAMPLE_SIZE, random_seed=RANDOM_SEED):
+    lines_count = count_lines(mesh_data_path)
```
lizgzil: should be `data_path`

scripts/create_mesh_sample.py
```diff
@@ -0,0 +1,43 @@
...
+    sample_line_indices = set(line_indices[:sample_size])
+
+    with open(data_path) as f:
+        for i, line in enumerate(f):
+            if i in sample_line_indices:
+                yield line
+
+def create_mesh_sample(mesh_data_path, sample_mesh_data_path):
+    with open(sample_mesh_data_path, "w") as f:
+        for item in yield_sample_data(mesh_data_path, sample_mesh_data_indices):
```
lizgzil: dont think `sample_mesh_data_indices` is an argument of `yield_sample_data`?

lizgzil:
sorry, was going through your latest commits in order so didnt see that youd already fixed those bits. All good 👍

nsorros:
> sorry, was going through your latest commits in order so didnt see that youd already fixed those bits. All good 👍

i fixed them after and just included the commit in the discussion, you are not going crazy :p

PR #29, opened 2020-09-29
Add script that evaluates mesh model on our grants

nsorros:
This script evaluates the current mesh model performance (tfidf-svm) in our grants data. To do that, Kirstin mapped the ICD codes that the annotators had used to MeSH.

The data collated the model predictions with the gold annotations, in the future we would like to use the model to produce the predictions so that we evaluate different models.

### DIFF
--- a/scripts/evaluate_mesh_on_grants.py
+++ b/scripts/evaluate_mesh_on_grants.py
@@ -1,0 +1,46 @@
+"""
+Evaluate MeSH model on grants annotated with MeSH
+
+The script works on an Excel file that contains 
+columns that follow the pattern below
+Disease MeSH#TAG_INDEX (ANNOTATOR)
+"""
+from argparse import ArgumentParser
+from pathlib import Path
+import pickle
+
+from sklearn.metrics import classification_report
+import pandas as pd
+
+
+def get_tags(data, annotator):
+    tag_columns = [col for col in data.columns if annotator in col]
+    tags = []
+    for _, row in data.iterrows():
+        row_tags = [tag for tag in row[tag_columns] if not pd.isnull(tag)]
+        tags.append(row_tags)
+    return tags
+
+def evaluate_mesh_on_grants(data_path, label_binarizer_path):
+    data = pd.read_excel(data_path)
+
+    with open(label_binarizer_path, "rb") as f:
+        label_binarizer = pickle.loads(f.read())
+
+    gold_tags = get_tags(data, "KW")
+    model_tags = get_tags(data, "NS")
+    Y = label_binarizer.transform(gold_tags)
+    Y_pred = label_binarizer.transform(model_tags)
+#    print(classification_report(Y, Y_pred, target_names=label_binarizer.classes_))
+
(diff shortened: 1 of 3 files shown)


scripts/evaluate_mesh_on_grants.py
```diff
@@ -0,0 +1,46 @@
...
+    data = pd.read_excel(data_path)
+
+    with open(label_binarizer_path, "rb") as f:
+        label_binarizer = pickle.loads(f.read())
+
+    gold_tags = get_tags(data, "KW")
+    model_tags = get_tags(data, "NS")
+    Y = label_binarizer.transform(gold_tags)
+    Y_pred = label_binarizer.transform(model_tags)
+#    print(classification_report(Y, Y_pred, target_names=label_binarizer.classes_))
```
nsorros, line 34: remove comment

nsorros:
[Evaluate disease mesh tagger on a sample of our grants tagged from annotators by mapping ICD codes to Mesh codes](https://trello.com/c/qSyVVCsb/1096-evaluate-disease-mesh-tagger-on-a-sample-of-our-grants-tagged-from-annotators-by-mapping-icd-codes-to-mesh-codes)

scripts/evaluate_mesh_on_grants.py
```diff
@@ -0,0 +1,47 @@
+"""
+Evaluate MeSH model on grants annotated with MeSH
+
+The script works on an Excel file that contains 
+columns that follow the pattern below
+Disease MeSH#TAG_INDEX (ANNOTATOR)
```
aCampello, line 6: Did Kirstin do all this manually? Impressive.

scripts/evaluate_mesh_on_grants.py
```diff
@@ -0,0 +1,47 @@
...
+    return tags
+
+def evaluate_mesh_on_grants(data_path, label_binarizer_path):
+    data = pd.read_excel(data_path)
+
+    with open(label_binarizer_path, "rb") as f:
+        label_binarizer = pickle.loads(f.read())
+
+    gold_tags = get_tags(data, "KW")
+    model_tags = get_tags(data, "NS")
```
aCampello, line 31: Nice aliases... 😉

aCampello (approved):
Looks good. What is the result? Maybe update the documentation with the current results?

scripts/evaluate_mesh_on_grants.py
nsorros: This is 2000 grants and she mapped previous annotators but 100% still impressive. Shout to @kirstinwilliamson 🥇

scripts/evaluate_mesh_on_grants.py
nsorros: Again shout out ot @kirstinwilliamson

nsorros:
> Looks good. What is the result? Maybe update the documentation with the current results?

Good point, let me see where best to add it. The result was 69% micro f1 which is impressive since the model performed 56% on the test set but it is partly explained by the non blind process i.e. Kirstin was looking at the model predictions and sometimes agreeing

PR #30, opened 2020-10-07
Calculate performance of MTI

nsorros:
Extend script that creates sample to create file in MTI input format.
Add script that reads both files i.e. the sample with the gold tags and the one output that MTI batch service sents back via email and calculates performance for a subset of mesh

### DIFF
--- a/scripts/create_mesh_sample.py
+++ b/scripts/create_mesh_sample.py
@@ -31,14 +31,26 @@ def yield_sample_data(data_path, sample_size=SAMPLE_SIZE, random_seed=RANDOM_SEE
             if i in sample_line_indices:
                 yield line
 
-def create_mesh_sample(mesh_data_path, sample_mesh_data_path):
+def create_mesh_sample(mesh_data_path, sample_mesh_data_path, mti_input_path):
+    if mti_input_path:
+        mti_f = open(mti_input_path, "w")
+
     with open(sample_mesh_data_path, "w") as f:
-        for item in yield_sample_data(mesh_data_path):
+        for i, item in enumerate(yield_sample_data(mesh_data_path)):
             f.write(item)
+
+            if mti_input_path:
+                uid = f"{i:08d}"
+                text = item["text"].encode("ascii", errors="ignore")
+                mti_f.write(f"{uid}|{text}\n")
+
+    if mti_input_path:
+        mti_f.close()
 
 argparser = ArgumentParser(description=__doc__.strip())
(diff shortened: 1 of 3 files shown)


nsorros:
[Calculate accuracy of MTI](https://trello.com/c/I6gengGr/1093-calculate-accuracy-of-mti)

nsorros:
@aCampello and @lizgzil just a reminder about this PR

scripts/create_mesh_sample.py
```diff
@@ -31,14 +31,26 @@ def yield_sample_data(data_path, sample_size=SAMPLE_SIZE, random_seed=RANDOM_SEE
...
+def create_mesh_sample(mesh_data_path, sample_mesh_data_path, mti_input_path):
+    if mti_input_path:
+        mti_f = open(mti_input_path, "w")
+
     with open(sample_mesh_data_path, "w") as f:
-        for item in yield_sample_data(mesh_data_path):
+        for i, item in enumerate(yield_sample_data(mesh_data_path)):
             f.write(item)
 
+            if mti_input_path:
```
aCampello, line 42: Perhaps a comment somewhere with a path to what is expected by MTI and what is output in the e mail? Or maybe in the README, add an example of command line/output for possible reproducibility?  Seems to me that it is as simple as a file with `unique_id|text`.

scripts/evaluate_mti.py
```diff
@@ -0,0 +1,56 @@
...
+            item = json.loads(line)
+            gold_tags.append(item["tags"])
+    return gold_tags
+
+def get_mti_tags(mti_output_path, mesh_tags):
+    mti_data = dict()
+    with open(mti_output_path) as f:
+        csvreader = csv.reader(f, delimiter="|")
+        for line in csvreader:
+            uid, tag, _, _ = line
```
aCampello, line 28: what is the other two returned fields?

aCampello (approved):
Looks good. Some very minor suggestions on documentation.

scripts/evaluate_mti.py
nsorros: cuid and score. the first is a unique identifier of the tag, score is not a probability but a custom scoring function. why?

scripts/create_mesh_sample.py
nsorros: 76c76dd

scripts/evaluate_mti.py
aCampello: I was just curious about why you are not using them. Makes sense.

PR #35, opened 2020-10-19
More text investigation

nsorros:
Some of our grants lack a Synopsis but contain a Lay Summary or Research question so this investigation is all about understanding how our model performs in those fields and whether we should apply it to all of them together or individuall or train a new model per field. The summary results below. What is decided is to use the same model and apply it to all available text concatenated which seems to perform the best.

Screenshot 2020-10-19 at 17 46 34

### DIFF
--- a/configs/2020.10.15.ini
+++ b/configs/2020.10.15.ini
@@ -1,0 +1,22 @@
+[DEFAULT]
+version=2020.10.14
+
+[preprocess]
+input=data/raw/science_tags_full_version.xlsx
+output=data/processed/science_grants_tagged_title_synopsis_question.jsonl
+text_cols=Title,Synopsis,Qu.
+
+[label_binarizer]
+data=data/processed/science_grants_tagged_title_synopsis.jsonl
+label_binarizer=models/label_binarizer.pkl
+
+[data]
+train_data_path=data/processed/science_grants_tagged_title_synopsis.jsonl
+test_data_path=data/processed/science_grants_tagged_title_synopsis_question.jsonl
+from_same_distribution=True
+
+[model]
+label_binarizer_path=models/label_binarizer.pkl
+model_path=models/tfidf-svm-2020.10.15.pkl
+approach=tfidf-svm
+parameters={'tfidf__min_df': 5, 'svm__estimator__class_weight': 'balanced', 'tfidf__ngram_range': (1,2)}
--- a/configs/2020.10.15.ini
+++ b/configs/2020.10.15.ini
@@ -1,5 +1,5 @@
 [DEFAULT]
-version=2020.10.14
+version=2020.10.15
 
 [preprocess]
 input=data/raw/science_tags_full_version.xlsx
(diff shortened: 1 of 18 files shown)


nsorros:
[Investigate performance difference when adding more text in science tagging](https://trello.com/c/hjceIyRl/1113-investigate-performance-difference-when-adding-more-text-in-science-tagging)

nsorros:
@aCampello @lizgzil no need of a thorough review, you can just skim read this one, it is mainly configs and results

configs/2020.10.15.ini
```diff
@@ -0,0 +1,22 @@
+[DEFAULT]
+version=2020.10.14
```
lizgzil, line 2: this should be 2020.10.15

docs/results.md
```diff
@@ -29,6 +29,22 @@ Version   | Approach    | Micro f1 | Description
 2020.06.1 | ensemble of tfidf and scibert | 0.75 | same as 2020.06.0 with threshold 0.3
 2020.06.2 | tfidf-adaboost | 0.63 | adaboost after tfidf
 2020.06.3 | tfidf-gboost | 0.66 | same as 2020.06.2 but with gradient boosting instead
+2020.10.0 | tfidf-svm | 0.72 | same as 2020.05.2 with train data including lay summary
+2020.10.1 | tfidf-svm | 0.72 | same as 2020.05.2 with train data including lay summary and research question
+2020.10.2 | tfidf-svm | 0.69 | same as 2020.05.2 with train data including lay summary but testing on title+synopsis
+2020.10.3 | tfidf-svm | 0.69 | same as 2020.05.2 with train data including lay summary and research question but testing on title+synopsis
```
lizgzil, line 35: doesnt seem to inc research q in the config `text_cols=Title,Synopsis,Lay Summary`

configs/2020.10.15.ini
nsorros: thanks for the thorough review, exactly what i needed

docs/results.md
nsorros: 139adcc also tested the results which stayed the same. this is probably because these files are not recreated if they exist so it was probably created from another config correctly

docs/results.md
```diff
@@ -29,6 +29,22 @@ Version   | Approach    | Micro f1 | Description
...
+2020.10.4 | tfidf-svm | 0.52 | same as 2020.05.2 but testing on titles only
```
lizgzil, line 36: I don't know how much it really matters since you have the configs but these descriptions are a bit confusing and inconsistent (i.e. here it doesnt say it was trained on title+synopsis+lay summary) , and i think they should just all follow the same format, e.g. like what you have for 2020.10.8. 'train on.... test on ... '

configs/2020.10.15.ini
nsorros: 78df98e

docs/results.md
nsorros: what i meant was that the train is the same as 2020.05.2 but i can make it as you say if it easier to read 👍

nsorros: 313416f

lizgzil (approved):
looks good :)

PR #36, opened 2020-10-20
Add script to evaluate performance by text length

nsorros:
This PR investigates the performance difference as we give more text to the model. Additionally it tests whether the first words are more important than a random pick of words and whether the effect observed is caused by the title.

The first graph is calculated on title + synopsis
performance_by_len

Whereas the second is calculated only on synopsis
performance_by_len_synopsis_only

We conclude that performance seems to peak at around 200 words and that the first 50 words contain far more important information about grant than a randon pick of words regardless of whether we include the title.


nsorros:
[Investigate performance difference when adding more text in science tagging](https://trello.com/c/hjceIyRl/1113-investigate-performance-difference-when-adding-more-text-in-science-tagging)

scripts/evaluate_performance_by_text_length.py
```diff
@@ -0,0 +1,92 @@
...
+import numpy as np
+
+
+def evaluate_model_on_different_lengths(model, X_test, Y_test, random_sample=False):
+    results = []
+    for word_len in range(50, 300, 50):
+        X_test_subset = []
+        Y_test_subset = []
+        for i in range(len(X_test)):
+            x_test_words = X_test[i].split()
```
lizgzil, line 29: I thought this evaluation would be on a random sentence/s rather than a random list of words. Not that a random list of words isn't also interesting, but it is a bit different since they are all jumbled up.
e.g.:
```
if random_sample:
    random.seed(42)
    start_i = random.randint(0,(len(x_test_words)-word_len))
    x_test_words = x_test_words[start_i:(start_i+word_len)]
```

lizgzil: i.e. it takes away your ability to say the first sentence holds the most info, since your test is also a structured sentence vs an unstructured sentence

scripts/evaluate_performance_by_text_length.py
```diff
@@ -0,0 +1,92 @@
...
+    ax.set_title("Performance of science tagger by text length")
+    ax.set_ylim(0.50)
+    ax.legend()
+
+    plt.savefig(figure_path)
+
+if __name__ == "__main__":
+    argparser = argparse.ArgumentParser(description=__doc__.strip())
+    argparser.add_argument("--label_binarizer", type=Path)
+    argparser.add_argument("--data_path", type=Path)
```
lizgzil, line 88: so youve run this with 2 data_paths - one to the data with just text=synopsis and one where text=title+synopsis?

scripts/evaluate_performance_by_text_length.py
nsorros: the structure of the sentence is lost though after when doing tfidf. i am not 100% convinced but it is also something i have been thinking, do you reckon i should try it and see if we get different results? we can also add it as a param and produce yet another graph for the different way to randomise

scripts/evaluate_performance_by_text_length.py
nsorros: exactly

scripts/evaluate_performance_by_text_length.py
nsorros: hey @lizgzil, do you want me to split on sentences as well in order to approve this?

lizgzil: no that's ok, will approve 👍

lizgzil (approved):

PR #37, opened 2020-10-21
Experiments with Team and Scheme for science tagging

nsorros:
This PR goal was to introduce an experiment on running tagging with additional meta columns such as Team and Scheme. This is so that we can evaluate the effects that those variables can have but also to prepare the ground for using additinal fields something also relevant for mesh tagging

To do so I had to
* change preprocessing to include those fields, this was a good opportunity to greatly simplify that module
* represent X as a DataFrame that not only contains list of texts but also those additional fields
* create approaches that combine text with meta features in sklearn

### DIFF
--- a/docs/results.md
+++ b/docs/results.md
@@ -45,6 +45,8 @@ Version   | Approach    | Micro f1 | Description
 2020.10.13 | tfidf-svm | 0.72 | same as 2020.5.2 but train on title+synopsis and test on title+synopsis+lay summary+research question
 2020.10.14 | tfidf-svm | 0.72 | same as 2020.5.2 but train on title+synopsis and test on title+synopsis+lay summary
 2020.10.15 | tfidf-svm | 0.71 | same as 2020.5.2 but train on title+synopsis and test on title+synopsis+research_question
+2020.10.16 | tfidf+onehot_team | 0.71 | similar to 2020.5.2 but we add team as one hot variable to the features. as data get shuffled differently it seems as the performance stays the same but it actually increases 1% compared to removing the team variable
+2020.10.17 | tfidf+onehot_scheme | 0.68 | similar to 2020.10.16 but this time we add scheme as one hot variable.
 
 ## 2020.02.0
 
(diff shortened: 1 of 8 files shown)


nsorros:
[Add non text fields to science tagging and record effect](https://trello.com/c/QL667u8t/1131-add-non-text-fields-to-science-tagging-and-record-effect)

grants_tagger/train.py
```diff
@@ -308,6 +337,7 @@ def train_and_evaluate(
     argparser.add_argument('-f', '--from_same_distribution', type=strtobool, default=False, help="whether train and test contain the same examples but differ in other ways, important when loading train and test parts of datasets")
     argparser.add_argument('--threshold', type=float, default=None, help="threhsold to assign a tag")
     argparser.add_argument('--y_batch_size', type=int, default=None, help="batch size for Y in cases where Y large. defaults to None i.e. no batching of Y")
+    argparser.add_argument('--x_format', type=str, default="List", help="format that will be used when loading the data. One of JSONL,DataFrame")
```
lizgzil, line 340: you say 'One of JSONL,DataFrame' but the default type is List?

docs/results.md
```diff
@@ -45,6 +45,8 @@ Version   | Approach    | Micro f1 | Description
 2020.10.13 | tfidf-svm | 0.72 | same as 2020.5.2 but train on title+synopsis and test on title+synopsis+lay summary+research question
 2020.10.14 | tfidf-svm | 0.72 | same as 2020.5.2 but train on title+synopsis and test on title+synopsis+lay summary
 2020.10.15 | tfidf-svm | 0.71 | same as 2020.5.2 but train on title+synopsis and test on title+synopsis+research_question
+2020.10.16 | tfidf+onehot_team | 0.71 | similar to 2020.5.2 but we add team as one hot variable to the features. as data get shuffled differently it seems as the performance stays the same but it actually increases 1% compared to removing the team variable
+2020.10.17 | tfidf+onehot_scheme | 0.68 | similar to 2020.10.16 but this time we add scheme as one hot variable.
```
lizgzil, line 49: what do you think is going on? the scheme has confused the model, needs to train for longer?

lizgzil: just a thought - but do you think maybe the scheme field needs cleaning, I wonder whether there is a lot of missing data or inconsistent scheme names? I see you do `processed_data.replace("No Data Entered", "").fillna("")` so maybe that would solve things.

lizgzil (approved):
looks good to me. Just had a thought about data cleaning

grants_tagger/train.py
nsorros: oups, good spot, will fix

docs/results.md
nsorros: you are wondering why performance drops instead of saying the same? this is what i was discussing in the slack channel saying that i find it unituitive when this happens even though it is a standard effect. scheme in our case does not carry much more information and possibly it only adds noise and dimensions to the dataset.

how do you suggest cleaning better? i am keen to further explore this if you have any ideas. atm scheme is treated as empty, of course we can deploy other imputation strategies but in terms of cleaning the data that exists, i think we are fine

lizgzil: oh yeh, I remember that conversation now. Yeh would assume it would stay the same so 3% down seems quite a bit!

I don't have any cleaning suggestions, just a thought.

grants_tagger/train.py
nsorros: 2ca0c15

PR #38, opened 2020-10-29
CNN model for disease mesh

nsorros:
* Extends evaluate ensemble to evaluate all models that have been deployed
* Adds ability to pass multiple thresholds to evaluate model to monitor different precision / recalls
* Adds predict and tag for new cnn model
* Adds configs and results with all experiments for cnn model so far


nsorros:

PR #39, opened 2020-11-03
Add tune threshold functionality

nsorros:
This PR implements tuning threshold for our tagger inspired from the paper https://doi.org/10.1016/j.patcog.2013.01.012 but modified to be efficient for MeSH. Modifications are:

- Tune threshold on a sample of the test set
- Try a fixed number of thresholds in the continuous range 0 to 1. For example if we want to try 5 thresholds they would be 0 0.2 0.4 0.6 0.8
- Init threshold value instead of starting from an unnecessarily bad init value which requires more iterations

### DIFF
--- a/grants_tagger/evaluate_model.py
+++ b/grants_tagger/evaluate_model.py
@@ -17,7 +17,7 @@ from grants_tagger.predict_mesh import predict_tfidf_svm, predict_cnn
 from grants_tagger.predict import predict_proba_ensemble_tfidf_svm_bert, load_model
 
 
-def predict(X_test, model_path, nb_labels=None, threshold=None, return_probabilities=False):
+def predict(X_test, model_path, nb_labels=None, threshold=0.5, return_probabilities=False):
     # comma indicates ensemble of more than one models
     if "," in model_path:
         model_paths = model_path.split(",")
--- a/grants_tagger/evaluate_model.py
+++ b/grants_tagger/evaluate_model.py
@@ -17,20 +17,26 @@ from grants_tagger.predict_mesh import predict_tfidf_svm, predict_cnn
 from grants_tagger.predict import predict_proba_ensemble_tfidf_svm_bert, load_model
 
 
-def predict(X_test, model_path, nb_labels, threshold):
+def predict(X_test, model_path, nb_labels=None, threshold=None, return_probabilities=False):
     # comma indicates ensemble of more than one models
     if "," in model_path:
         model_paths = model_path.split(",")
         Y_pred_proba = predict_proba_ensemble_tfidf_svm_bert(X_test, model_paths)
-        Y_pred = Y_pred_proba > threshold
+        if return_probabilities:
+            Y_pred = Y_pred_proba
+        else:
+            Y_pred = Y_pred_proba > threshold
     elif "disease_mesh_cnn" in model_path:
-        Y_pred = predict_cnn(X_test, model_path, threshold)
+        Y_pred = predict_cnn(X_test, model_path, threshold, return_probabilities)
     elif "disease_mesh_tfidf" in model_path:
-        Y_pred = predict_tfidf_svm(X_test, model_path, nb_labels, threshold)
+        Y_pred = predict_tfidf_svm(X_test, model_path, nb_labels, threshold, return_probabilities)
     else:
         model = load_model(model_path)
         Y_pred_proba = model.predict_proba(X_test)
-        Y_pred = Y_pred_proba > threshold
+        if return_probabilities:
+            Y_pred = Y_pred_proba
+        else:
+            Y_pred = Y_pred_proba > threshold
     return Y_pred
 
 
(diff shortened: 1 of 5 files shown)


nsorros:
[Tune threshold for each mesh term or label term independently to maximise f1](https://trello.com/c/wWyX0lzP/1136-tune-threshold-for-each-mesh-term-or-label-term-independently-to-maximise-f1)

grants_tagger/tune_threshold.py
```diff
@@ -0,0 +1,134 @@
...
+
+    return optimal_thresholds
+
+
+def tune_threshold(data_path, model_path, label_binarizer_path, thresholds_path, sample_size=None, nb_thresholds=None, init_threshold=None):
+    with open(label_binarizer_path, "rb") as f:
+        label_binarizer = pickle.loads(f.read())
+
+    X, Y, _ = load_data(data_path, label_binarizer)
+    _, X_test, _, Y_test = train_test_split(X, Y, random_state=42)
```
lizgzil, line 98: is this the same split you used when training the model? if it's not, I guess it might overfit your thresholds to the training data?

grants_tagger/evaluate_model.py
```diff
@@ -17,20 +17,26 @@
 from grants_tagger.predict import predict_proba_ensemble_tfidf_svm_bert, load_model
 
 
-def predict(X_test, model_path, nb_labels, threshold):
+def predict(X_test, model_path, nb_labels=None, threshold=None, return_probabilities=False):
```
lizgzil, line 20: perhaps `threshold` should be `thresholds`?

grants_tagger/tune_threshold.py
```diff
@@ -0,0 +1,134 @@
...
+    optimal_threshold_i = current_threshold_i
+
+    Y_pred = Y_pred_proba > candidate_thresholds
+    Y_pred = csr_matrix(Y_pred)
+    cm = multilabel_confusion_matrix(Y_test, Y_pred)
+
+    if nb_thresholds:
+        candidate_thresholds_i = [t/nb_thresholds for t in range(1, nb_thresholds)]
+    else:
+        candidate_thresholds_i = Y_pred_proba[:, label_i]
```
lizgzil, line 38: Might be worth `np.unique(Y_pred_proba[:, label_i])`?

grants_tagger/tune_threshold.py
nsorros: this is a good point, i think i might have another function that splits but essentially that's what it does under the hood. i will check and replace

grants_tagger/tune_threshold.py
nsorros: good idea

grants_tagger/evaluate_model.py
nsorros: tricky one as you can pass in both a value and a list, not sure what the right name in these cases

lizgzil:
also `make test` passed for me 👍

nsorros:
> also `make test` passed for me 👍

that's some good news

grants_tagger/tune_threshold.py
nsorros: 8b1752b

grants_tagger/tune_threshold.py
nsorros: 59aa03e

grants_tagger/evaluate_model.py
```diff
@@ -17,20 +17,26 @@
...
+        if return_probabilities:
+            Y_pred = Y_pred_proba
+        else:
+            Y_pred = Y_pred_proba > threshold
     elif "disease_mesh_cnn" in model_path:
-        Y_pred = predict_cnn(X_test, model_path, threshold)
+        Y_pred = predict_cnn(X_test, model_path, threshold, return_probabilities)
     elif "disease_mesh_tfidf" in model_path:
-        Y_pred = predict_tfidf_svm(X_test, model_path, nb_labels, threshold)
+        Y_pred = predict_tfidf_svm(X_test, model_path, nb_labels, threshold, return_probabilities)
```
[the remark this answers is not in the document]
nsorros: actually this is a bug well spotted. added a test as well. fixed in 3516143

grants_tagger/evaluate_model.py
```diff
@@ -17,20 +17,26 @@
...
+def predict(X_test, model_path, nb_labels=None, threshold=None, return_probabilities=False):
     # comma indicates ensemble of more than one models
     if "," in model_path:
         model_paths = model_path.split(",")
         Y_pred_proba = predict_proba_ensemble_tfidf_svm_bert(X_test, model_paths)
-        Y_pred = Y_pred_proba > threshold
+        if return_probabilities:
+            Y_pred = Y_pred_proba
+        else:
+            Y_pred = Y_pred_proba > threshold
```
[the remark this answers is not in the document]
nsorros: good point, will default to None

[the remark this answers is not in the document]
nsorros: 2a49f37

lizgzil (approved):

PR #40, opened 2020-11-05
Evaluate model

nsorros:
This PR introduces some changes to evaluate, predict and tag grants in light of the new neural model for disease mesh. There are two main changes here and a couple of smaller ones. The two main changes are

- evaluate ensemble becomes evaluate model and can now evaluate any deployed model
- evaluate model can accept multiple thresholds and in that case it calculates precision and recall for each threshold, a process we follow to set up a static threshold for all tags

There are a couple of things that are not optimal which I plan to fix in the future. Those are

- the model path directs which model is loaded and which predict function is called. it would be ideal if the models behaved in a more similar way after loading and that the path contained some meta information about how to load them. this way the code would be more uniformal and the complexity could be hidden behind a load function that takes instruction from a meta file
- the two tagging projects have separate functions but very similar code. in fact the code has converged nicely to a point where it is now easy to combine the functionality so that irrespective of whether it is the disease mesh tags or science tags the codebase is mostly the same especially in regards to predict and tag which should not differ

### DIFF
--- a/grants_tagger/predict.py
+++ b/grants_tagger/predict.py
@@ -21,7 +21,7 @@ DEFAULT_LABELBINARIZER_PATH = os.path.join(FILEPATH, '../models/label_binarizer.
 
 
 def load_model(model_path):
-    if 'pkl' in model_path[-4:]:
+    if model_path.endswith('.pkl'):
         with open(model_path, "rb") as f:
             model = pickle.loads(f.read())
             return model
--- a/grants_tagger/predict.py
+++ b/grants_tagger/predict.py
@@ -39,7 +39,7 @@ def predict_proba_ensemble_tfidf_svm_bert(X, model_paths):
         Y_pred_proba_model = model.predict_proba(X)
         Y_pred_proba.append(Y_pred_proba_model)
 
-    Y_pred_proba = np.sum(Y_pred_proba) / len(model_paths)
+    Y_pred_proba = np.array(Y_pred_proba).sum(axis=0) / len(model_paths)
     return Y_pred_proba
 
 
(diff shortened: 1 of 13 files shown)


nsorros:

grants_tagger/predict.py
```diff
@@ -20,25 +20,37 @@
 DEFAULT_LABELBINARIZER_PATH = os.path.join(FILEPATH, '../models/label_binarizer.pkl')
 
 
+def load_model(model_path):
+    if 'pkl' in model_path[-4:]:
```
aCampello, line 24: `model_path.endswith('.pkl')` will also do.

grants_tagger/predict_mesh.py
```diff
@@ -4,30 +4,73 @@
...
+            Y_pred_proba.append(Y_pred_proba_batch)
+        Y_pred_proba = vstack(Y_pred_proba)
+        return Y_pred_proba
+    else:
+        Y_pred = model.predict(X_vec)
+        return Y_pred
+
+
+def predict_mesh_tags(X, model_path, label_binarizer_path,
+                      probabilities=False, threshold=0.5,
```
aCampello, line 60: How does this `probabilities` compare with `probability` below? Is it the same (in which case you should call the function with `probability=probabilities` instead of `probability=True` ?

aCampello (approved):
Looks good to me. Some minor points. Agree with the point in the PR description that the code is getting a bit complicated with lots of `ifs` for different behaviours and different models and would be good to unify.

grants_tagger/predict.py
nsorros: 4f05ae3

PR #41, opened 2020-11-05
Neural model for MeSH results and configs

nsorros:
This PR adds configs and results for all experiments since CNN started working well (above 64%). As such the experiments recorded are variations on the first config that performs well and so far none of the variations has not exceeded that performance which builds on our confidence that parameters are well picked. We can not do an exhaustive search of params as training is expensive.

**Note**
I do not include results on evaluating the models against our grants as this requires some transformation of the data because of the data was initially shared. I plan to add that in a future PR


nsorros:

docs/mesh_results.md
```diff
@@ -10,6 +10,13 @@ MTI      | 0.49
 Version   | Approach    | Micro f1 in Pubs | Micro f1 in Grants | Description
 --------- | ----------- | ---------------- | ------------------ | -------
 2020.07.0 | tfidf-sgd   | 0.56             | 0.69               | tfidf-svm, bigrams, regularisation of 1e-9, removal of words appearing less than 5 times
+2020.9.0  | cnn         | 0.64             |                    | 1e-3 lr, 0.1 dr, 1e-7 l2, 4 layers, attention
+2020.9.1  | cnn         | 0.63             |                    | 1e-3 lr, 0.1 dr, **1e-8 l2**, 4 layers, attention
+2020.9.2  | cnn         | 0.64             |                    | 1e-3 lr, 0.1 dr, 1e-7 l2, 4 layers, ~~attention~~
+2020.9.3  | cnn         | 0.57             |                    | 1e-3 lr, 0.1 dr, 1e-7 l2, **0 layers**, attention
+2020.9.4  | cnn         | 0.62             |                    | 1e-3 lr, **0 dr**, 1e-7 l2, 4 layers, attention
```
aCampello, line 17: We don't have any result on grants?

aCampello (approved):
LGTM - I didn't verify carefully all config files but the results make sense. Why don't we have results on grants?

docs/mesh_results.md
nsorros: I will add separately. The problem is the format of the data that needs a bit of preprocessing to run the results. Will do very soon

PR #44, opened 2020-11-19
Convert threshold to float before evaluating

### LINKED ISSUE #43: Threshold is str instead of float in evaluate model
Evaluate model accepts thresholds as an str argument to allow for multiple comma separated thresholds but does not convert to float before evaluating which throws an error.

nsorros:
Fixes #43

### DIFF
--- a/grants_tagger/evaluate_model.py
+++ b/grants_tagger/evaluate_model.py
@@ -81,5 +81,7 @@ if __name__ == '__main__':
 
     # comma indicates multiple threshold to evaluate against
     if "," in threshold:
-        threshold = threshold.split(",")
+        threshold = [float(t) for t in threshold.split(",")]
+    else:
+        threshold = float(threshold)
     evaluate_model(models, data, label_binarizer, threshold)


nsorros:
@aCampello @lizgzil

grants_tagger/evaluate_model.py
```diff
@@ -81,5 +81,7 @@ def evaluate_model(model_path, data_path, label_binarizer_path, threshold):
 
     # comma indicates multiple threshold to evaluate against
     if "," in threshold:
-        threshold = threshold.split(",")
+        threshold = [float(t) for t in threshold.split(",")]
+    else:
+        threshold = float(threshold)
```
lizgzil, line 86: not sure how likely this scenario is, so feel free to ignore, but if someone inputs `threshold = '0.4, 0.5, '` (i.e. a comma at the end) this will break. Just thinking this might come up if you are reading from a file maybe?

lizgzil (approved):
I'll approve, but you might think my edge case scenario is worth accounting for

grants_tagger/evaluate_model.py
nsorros: interesting but i am tempted to say that this case is not supported for now

PR #48, opened 2020-12-14
Add CLI tool and remove unused scripts

### LINKED ISSUE #22: Create entrypoint for grants_tagger
Copy form jeff comment in #21 
> If you look here https://github.com/wellcometrust/rori/blob/master/setup.py#L23 at how we have entry_points set up on rori, it means that when you install grants_tagger onto a system, it will make an alias in /usr/bin/ so that you can run the package as a CLI tool, then you don't need to wrap the library in an outer python script to run it.
You can just to <entrypoint_name> --<arg> <value> without having to invoke python or anything.
Isn't necessary now, but worth noting for libs like this which have a CLI interface defined.

nsorros:
Fixes #22 

Creates CLI tools grants_tagger with commands
```
grants_tagger COMMAND

Commands
* train
* pretrain
* preprocess (subcommands `bioasq-mesh` and `wellcome-science`)
* predict
* tag
* tune (subcommands `thresholds` and `params`)
* download
* explain
* evaluate
```


grants_tagger/preprocess.py
```diff
@@ -34,14 +34,24 @@ def yield_preprocess_data(
     # cols = ['Grant Team', 'ERG', 'Lead Applicant', 'Organisation', 'Scheme', 'Title', 'Synopsis', 'Lay Summary', 'Qu.']
     processed_data = data.drop_duplicates(subset=["Grant_ID", "Sciencetags"])
     processed_data = processed_data.dropna(subset=["Synopsis"])
+    processed_data["intersection"] = processed_data["intersection"].str.replace("19: Maternal, Child", "19: Maternal Child")
+    processed_data["intersection"] = processed_data["intersection"].str.replace("24: Data Science, Computational", "24: Data Science Computational")
```
lizgzil, line 38: when i do `make test` the `test_preprocess.py` tests fail. This seems to be because the 'intersection' key isnt in the data for the tests? i dont think `tagger 1 only` is in them either. I wonder whether it's worth doing a processed_data.get('intersection') or should intersection just always be in the data being processed?

grants_tagger/preprocess.py
```diff
@@ -34,14 +34,24 @@ def yield_preprocess_data(
...
     processed_data = processed_data.groupby('Grant_ID').agg({
         'Sciencetags': lambda x: ",".join(x),
         'Title': lambda x: x.iloc[0],
         'Synopsis': lambda x: x.iloc[0],
         'Lay Summary': lambda x: x.iloc[0],
         'Qu.': lambda x: x.iloc[0],
         'Scheme': lambda x: x.iloc[0],
-        'Team': lambda x: x.iloc[0]
+        'Team': lambda x: x.iloc[0],
+        'Tagger1_tags': lambda x: x.iloc[0]
```
lizgzil, line 54: also i think you might want to add 'intersection' and 'tagger 1 only' to the output?

lizgzil (commented):
make tests fail on test_preprocess.py :(

grants_tagger/preprocess.py
nsorros: what do you mean? i add them here, as there are multiple duplicate rows in the data. Kirstin broke the data to one row per tag for the tags with all other information duplicated. so all i am doing here is choosing one of the rows.

nsorros:
> make tests fail on test_preprocess.py :(

i am aware, will fix today 👍

grants_tagger/preprocess.py
nsorros: should it always be? that's a good question, probably not. let me think about it.

grants_tagger/preprocess.py
lizgzil: ah yeh sorry I didnt notice that you use them to create "Tagger1_tags"

lizgzil (approved):
tests pass 👍

nsorros:
> tests pass 👍

perfect, i will give them an opportunity to pass here as well and merge

PR #49, opened 2020-12-15
Increase test coverage

### LINKED ISSUE #19: Add tests for predict
Atm predict loads the models and binarizer outside the context of any function so tests would fail as they do not have access to the models. I wonder what is the best way to add tests here? Should I somehow include the models somewhere where the tests can access them, should i load the models inside the function? @jdu any thoughts?

nsorros:
Adds tests and fixes bugs along the way

Fixes #19


lizgzil (approved):
tests pass and the code looks good

PR #56, opened 2021-02-11
Incremental training

nsorros:
This PR reintroduces the idea of incremental learning which we first wanted to use to train our Tfidf-SVM baseline.

This time we use our CNN model which leverages `tf.data.Dataset` to implement incremental learning. Note that tensorflow also allows for a generator to be passed in but in that case functionally such as batching, shuffling etc would need to be implemented whereas now it comes for free.

Note that `KerasVectorizer` and `MultiLabelBinarizer` allow for a generator to be passed in so training them efficiently is relatively easy. One thing we need to work around, as is often in this project, is sparse labels. In particular the dataset converts to dense as it returns data but this means that test_data need to be converted back to sparse to fit into memory.

### DIFF
--- a/grants_tagger/utils.py
+++ b/grants_tagger/utils.py
@@ -4,6 +4,7 @@ import json
 
 from sklearn.model_selection import train_test_split
 from sklearn.metrics import f1_score
+import tensorflow as tf
 import pandas as pd
 import numpy as np
 
@@ -32,6 +33,18 @@ def load_data(data_path, label_binarizer=None, X_format="List"):
 
     return texts, tags, meta
         
+
+def load_tags(data_path, label_binarizer=None):
+    tags = []
+    with open(data_path) as f:
+        for line in f:
+            item = json.loads(line)
+            tags.append(item["tags"])
+
+    if label_binarizer:
+        tags = label_binarizer.transform(tags)
+    return tags
+
 
 def load_train_test_data(
         train_data_path, label_binarizer,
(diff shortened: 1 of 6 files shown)


nsorros:
[Train CNN on entire MeSH](https://trello.com/c/GTAM7HL6/1174-train-cnn-on-entire-mesh)

nsorros:
Just a reminder about this PR @lizgzil @aCampello

requirements.txt
```diff
@@ -1,4 +1,4 @@
 scikit-multilearn
 arff
 python-dateutil<2.8.1,>=2.1
--e git+git://github.com/wellcometrust/WellcomeML.git@ca5d1c38f28a50aacdefa5b5b47647389194e826#egg=wellcomeml[deep-learning]
+-e git+git://github.com/wellcometrust/WellcomeML.git@14668e8650516363453a85bc2a5546cfe8811951#egg=wellcomeml[deep-learning]
```
aCampello, line 4: Do you not want to use the pypi version?

aCampello (approved):
Looks good. Wondering why you are calling 'incremental learning' is it because you are loading the data at every batch ? I thought about something like: when you get more annotated data, you can pick the model from where you stopped and continue training..

grants_tagger/utils.py
```diff
@@ -95,3 +68,78 @@ def calc_performance_per_tag(Y_true, Y_pred, tags):
...
+        text_encoded = tokenizer.transform(texts)
+        tags_encoded = label_binarizer.transform(tags)
+
+        if sparse_labels:
+            tags_encoded = tags_encoded.todense() # returns matrix
+            tags_encoded = np.asarray(tags_encoded)
+        
+        return text_encoded, tags_encoded
+
+    def data_gen():
```
lizgzil, line 96: it would be nice for some comments on this or a function rename, since it isnt just a data generator but a kind of windowed data generator?

requirements.txt
nsorros: haha, because I so often submit a feature and fix I keep pointing to the latest commit to develop new features and use the latest published when I "publish" a new version

grants_tagger/utils.py
nsorros: I can add comments for sure, guilty on that, need to write more docstrings 😬 

this is actually a callable generator that returns one example at a time. why callable? because when it returns all elements tensorflow needs a way to restart so it basically calls it again, otherwise it would be able to iterate only once.

```
def data_gen():
    for i in range(5):
        yield I

non_callable_generator = data_gen()
callable_generator = data_gen
```

nsorros:
> Looks good. Wondering why you are calling 'incremental learning' is it because you are loading the data at every batch ? I thought about something like: when you get more annotated data, you can pick the model from where you stopped and continue training..

I was looking for a good name and settled with incremental learning, the older name was online_learning. I am trying to describe situation where you can update an algorithm with some examples which not algorithms allow, in this case you can leverage "incremental learning" to not load all data into memory and give them one by one. Possibly another name could be related to the memory efficiency thing.

see also: https://scikit-learn.org/0.15/modules/scaling_strategies.html

PR #58, opened 2021-02-23
Update WellcomeML to 2021.2.1

nsorros:
Updating to WellcomeML contains transition to spacy v3 and
the rewrite of BertClassifier among other things.

~Spacy v3 meant that we needed to update scispacy but as it turns out
this was with its own issues https://github.com/allenai/scispacy/issues/324
so I am postponing that until the issue is resolved and disabling the tests~

BertClassifier switched from referring to pretrained models with short names
to referencing the hugging face names so I am updating the tests here.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,7 @@ install:
 
 script:
   - make test
-  # - make test_scispacy
+  - make test_scispacy
 
 branches:
   only:
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,7 @@ install:
 
 script:
   - make test
-  - make test_scispacy
+  # - make test_scispacy
 
 branches:
   only:
--- a/tests/test_train.py
+++ b/tests/test_train.py
@@ -76,7 +76,7 @@ def test_create_scibert_svm():
 def test_create_bert():
     model = create_model('bert')
     assert isinstance(model, BertClassifier)
-    assert model.pretrained == 'bert'
+    assert model.pretrained == 'bert-base-uncased'
 
 
 def test_create_scibert():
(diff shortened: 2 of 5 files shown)


tests/test_train.py
```diff
@@ -76,7 +76,7 @@ def test_create_scibert_svm():
 def test_create_bert():
     model = create_model('bert')
     assert isinstance(model, BertClassifier)
-    assert model.pretrained == 'bert'
+    assert model.pretrained == 'bert-base-uncased'
```
aCampello, line 79: why uncased? smaller?

.travis.yml
```diff
@@ -8,7 +8,7 @@ install:
 
 script:
   - make test
-  - make test_scispacy
+  # - make test_scispacy
```
aCampello, line 11: you don't need scispacy anymore?

aCampello (approved):
LGTM - I don't quite understand what the new pipe does, can you clarify? 

```    nlp.add_pipe("scispacy_linker", config={"resolve_abbreviations": True, "linker_name": "mesh"})```

tests/test_train.py
nsorros: I was always using the uncased model, have you been using the cased on?

.travis.yml
nsorros: I think a later commit restores it but if not I will amend. The reason for removing is in the PR description.

nsorros:
> LGTM - I don't quite understand what the new pipe does, can you clarify?
> 
> ` nlp.add_pipe("scispacy_linker", config={"resolve_abbreviations": True, "linker_name": "mesh"})`

Same as the old, just updated to follow spacy v3 and resolved an issue on the scispacy library, essentially renaming the variable `name` to `linker_name`

PR #60, opened 2021-03-03
Add classes for production models

### LINKED ISSUE #23: Use WellcomeVotingClassifier for science tagging
Currently the latest version uses custom code to combine the probabilities from the two pretrained models whereas it would be nice to make use of the WellcomeVotingClassifier to wrap the models into one model so that the interaction is easier.

### LINKED ISSUE #24: Create a model class for the baseline TFIDF-SVM model of mesh
Currently the model needs to batch the y vector since it does not easily fit into memory. It would be more sklearn like if we can wrap this functionality into a model class so that we can run predict and it did the loading, predicting combing

nsorros:
The main motivation for this PR was that pieces of the implementation that made production models happen was living in different scripts such as train, predict and evaluate. Those were custom adjustments that are not general enough to go to WellcomeML so they ended up in a semi organised way in the project. As progressively these adjustments are quite specific to the some models and in particular those that went to production, the logical thing to do was to introduce custom classes for the production models that hide away the complexity and expose a clean sklearn API such as WellcomeML. In order for this to happen a lot of code needed to move around and some needed to be rewritten hence the big size which as you can see does not introduce a lot of new lines but does move around a lot.

Production model adjustments that moved into classes
* y_batch_size was a parameter that was controlling batching in the Y vector in mesh tfidf svm and is now part of that model
* save and load is implemented inside the production classes
* the ensemble model has its own class which makes use of WellcomeVotingClassifier
* from_same_distribution is removed, only used in two configs and added non intuitive complexity

Other changes that are worth highlighting
* tests now use fixtures to reduce duplication
* load_train_test_data returns a list or generator and `data_format` in train controls that
* models save a meta file which can be used in the future to restore variables and know which model will be loaded
* scispacy does not need a separate installation since the latest WellcomeML uses spacy v3

Fixes #23
Fixes #24

### DIFF
--- a/grants_tagger/models.py
+++ b/grants_tagger/models.py
@@ -308,6 +308,9 @@ class MeshTfidfSVM():
         if not hasattr(self, 'classifier'):
             self._init_classifier()
 
+        # TODO: Currently Y is expected to be sparse, otherwise predict does not
+        # work, add a check and warn user.
+
         print(f"Creating {self.model_path}")
         Path(self.model_path).mkdir(exist_ok=True)
         print("Fitting vectorizer")
--- a/grants_tagger/models.py
+++ b/grants_tagger/models.py
@@ -124,7 +124,7 @@ class MeshCNN():
 
         X: list of texts
         vectorizer: vectorizer class that implements transform which transforms texts to integers
-        Y: 2d numpy array that represents targets (tags) assigned.
+        Y: 2d numpy array or sparse csr_matrix that represents targets (tags) assigned.
 
         If Y is missing, for example when called by predict, yield_data yields only X vectorized
         """
@@ -209,8 +209,8 @@ class MeshCNN():
 
     def fit(self, X, Y):
         """
-        X: list, np.array or generator of texts
-        Y: 2d np.array of tags assigned 
+        X: list or generator of texts
+        Y: 2d numpy array or sparse csr_matrix or generator of 2d numpy array of tags assigned 
 
         If X is a generator it need to be callable i.e. return 
         the generator by calling it X_gen = X(). This is so
@@ -347,8 +347,8 @@ class MeshTfidfSVM():
 
     def fit(self, X, Y):
         """
-        X: np.array or list of texts
-        Y: 2d np.array of tags assigned
+        X: list of texts
+        Y: sparse csr_matrix of tags assigned
         """
         if not hasattr(self, 'vectorizer'):
             self._init_vectorizer()
--- a/grants_tagger/models.py
+++ b/grants_tagger/models.py
@@ -356,7 +356,9 @@ class MeshTfidfSVM():
         return self._predict(X, return_probabilities=True)
 
     def save(self, model_path):
-        # model saved during fit
+        if model_path != self.model_path:
+            print(f"{model_path} is different from self.model_path {self.model_path}. This will result in model and meta.json be saved in different paths")
+
         meta = {
             "name": "MeshTfidfSVM",
             "approach": "mesh-tfidf-svm",
--- a/grants_tagger/models.py
+++ b/grants_tagger/models.py
@@ -32,6 +32,15 @@ class ApproachNotImplemented(Exception):
 
 
 def get_params_for_component(params, component):
+    """
+    Returns a dictionary of all params for one component defined
+    in params in the form component__param: value
+
+    e.g.
+    >> params = {"vec__min_df": 1, "clf__probability": True}
+    >> get_params_for_component(params, "vec")
+    {"min_df": 1}
+    """
     component_params = {}
     for k, v in params.items():
         if k.startswith(component):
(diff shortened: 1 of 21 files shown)


nsorros:
[Introduce classes for production models](https://trello.com/c/0keWvl2S/1188-introduce-classes-for-production-models)

nsorros:
@aCampello @lizgzil feel free to skim read and see if you find anything weird, I understand it is not realistic to review the entire thing. hopefully tests cover most of the code anyway.

possibly the most important thing to review is the production models which are in the models file.

grants_tagger/models.py
```diff
@@ -0,0 +1,604 @@
...
+        random_seed: int, default 42. Random seed that controls reproducibility.
+        """
+        self.threshold = threshold
+        self.batch_size = batch_size
+        self.shuffle = shuffle
+        self.buffer_size = buffer_size
+        self.data_cache = data_cache
+        self.random_seed = random_seed
+
+    def _yield_data(self, X, vectorizer, Y=None):
```
aCampello, line 121: I have the feeling that there is an opportunity to move this to WellcomeML, as we use generators there too... Let's keep it in mind.

grants_tagger/models.py
```diff
@@ -0,0 +1,604 @@
...
+        model = BertClassifier()
+    elif approach == 'scibert':
+        model = BertClassifier(pretrained='scibert')
+    elif approach == 'classifierchain-tfidf-svm':
+        model = Pipeline([
+            ('tfidf', TfidfVectorizer(
+                stop_words='english', max_df=0.95,
+                min_df=0.0, ngram_range=(1,1)
+            )),
+            ('svm', ClassifierChain(classifier=SVC(kernel='linear', probability=True)))
```
aCampello, line 462: These parameters for SVC - did you get it after doing some sort of grid-search? Not related to this PR specifically, but wondering whether changing `C` and kernel to `RBF` will give you a boost

aCampello (approved):
I have to be honest that I didn't read it very carefully. I focused on the `models.py` file but it looks good at hindsight. 

One thing I wondered when reading is whether you actually need all those for production, i.e., once you decide on one of the models, the most performant one, we will probably stick to it.

setup.py
```diff
@@ -32,11 +32,12 @@ def package_files(directory):
         'scikit-learn==0.23.2',
         'nltk',
         'matplotlib',
-        'wellcomeml[deep-learning]==2021.2.1',
+        'wellcomeml[deep-learning]==1.0.2',
```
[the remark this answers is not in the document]
nsorros: this is intentional, I tend to point requirements.txt to the latest WellcomeML for dev purposes.

grants_tagger/models.py
nsorros: yes 100% basically WellcomeML currently expects tensorflow dataset instead of generators but there is no reason why we should not support both

grants_tagger/models.py
nsorros: I think RBF does not have probability or you need to do something else to get a probability. These params are just init values, params are also set in the init files for every experiment. So possibly it would be cleaner not to set any param and to always override all params using init files. I will resolve that when I move to DVC

nsorros:
> I have to be honest that I didn't read it very carefully. I focused on the `models.py` file but it looks good at hindsight.
> 
> One thing I wondered when reading is whether you actually need all those for production, i.e., once you decide on one of the models, the most performant one, we will probably stick to it.

Good point, worth noting that these are 3 models and 2 of them are for difference data i.e. science tags and mesh tags. So only for mesh there are two approaches which have made it to production. I guess my thinking for those is to record what was done to make an approach work in case others or ourselves find it useful in the future.

nsorros:
Thanks for the review @aCampello, you are a hero 🥇

grants_tagger/models.py
```diff
@@ -0,0 +1,604 @@
...
+from sklearn.preprocessing import Normalizer, OneHotEncoder, FunctionTransformer
+from scipy import sparse as sp
+import tensorflow as tf
+import numpy as np
+
+from wellcomeml.ml import BiLSTMClassifier, CNNClassifier, KerasVectorizer, SpacyClassifier, BertVectorizer, BertClassifier, Doc2VecVectorizer, Sent2VecVectorizer, WellcomeVotingClassifier
+
+
+class ApproachNotImplemented(Exception):
+    pass
```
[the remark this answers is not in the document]
nsorros: how would that work? can you expand?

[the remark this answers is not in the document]
aCampello: Sure! It's really just bells and whistles, but if you create a init function inside the class, you can define what is the standard error message

```python
 def __init__(self, message="The ML approach is not implemented"):
    self.message = message
    super().__init__(message)
```

Check here (https://docs.python.org/3/tutorial/errors.html#user-defined-exceptions)

PR #61, opened 2021-03-23
Transformers tokenizer experiment for science tags

nsorros:
Add experiments for using Transformers tokenizer for science tagging

### DIFF
--- a/configs/2020.04.3.ini
+++ b/configs/2020.04.3.ini
@@ -16,6 +16,6 @@ test_data_path=
 
 [model]
 label_binarizer_path=models/label_binarizer.pkl
-model_path=models/cnn-2020.04.3
+# model_path=models/cnn-2020.04.3
 approach=cnn
 parameters=
--- a/configs/2021.03.0.ini
+++ b/configs/2021.03.0.ini
@@ -1,0 +1,21 @@
+[DEFAULT]
+version=2021.02.0
+
+[preprocess]
+input=data/raw/science_tags_full_version.xlsx
+output=data/processed/science_grants_tagged_title_synopsis.jsonl
+text_cols=
+
+[label_binarizer]
+data=data/processed/science_grants_tagged_title_synopsis.jsonl
+label_binarizer=models/label_binarizer.pkl
+
+[data]
+train_data_path=data/processed/science_grants_tagged_title_synopsis.jsonl
+test_data_path=
+
+[model]
+label_binarizer_path=models/label_binarizer.pkl
+approach=cnn
+parameters={"vec__tokenizer_library": "transformers", "vec__sequence_length": 400, "vec__vocab_size": 5000,"cnn__nb_epochs": 20}
+
--- a/configs/2021.03.0.ini
+++ b/configs/2021.03.0.ini
@@ -17,5 +17,5 @@ test_data_path=
 [model]
 label_binarizer_path=models/label_binarizer.pkl
 approach=cnn
-parameters={"vec__tokenizer_library": "transformers", "vec__sequence_length": 400, "vec__vocab_size": 5000,"cnn__nb_epochs": 20}
+parameters={"vec__tokenizer_library": "transformers", "vec__sequence_length": 400, "vec__vocab_size": 30000,"cnn__nb_epochs": 20}
 
--- a/configs/2021.03.0.ini
+++ b/configs/2021.03.0.ini
@@ -1,5 +1,5 @@
 [DEFAULT]
-version=2021.02.0
+version=2021.03.0
 
 [preprocess]
 input=data/raw/science_tags_full_version.xlsx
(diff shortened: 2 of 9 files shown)


nsorros:
[Train CNN with TransformerTokenizer that will result in smaller vocabulary and thus smaller model](https://trello.com/c/ixcKfgnI/1167-train-cnn-with-transformertokenizer-that-will-result-in-smaller-vocabulary-and-thus-smaller-model)

nsorros:
I noticed that tokenisation is stochastic so results vary when you run multiple times. I have opened an issue to the library to see what is the best way to resolve this and make it more reproducible. https://github.com/huggingface/tokenizers/issues/668

This is a problem in CNN not in tfidf-svm for some reason. The range of values goes from 0.48 to 0.58 for a vocab size of 5K whereas for 30K seems to reduce to 0.54 to 0.58.

nsorros:
This is ready for review @aCampello @lizgzil, the reproducibility problem was resolved by setting up the seed in tensorflow. It is still weird that transformers tokenizer is producing different results each time but it does not seem to affect reproducibility. I did not manage to get the 0.60 that I was able to when I first run but got close to 0.58 and called it a day. the variance is still very high for CNN which is possibly something to improve in the future.

configs/2020.04.3.ini
```diff
@@ -16,6 +16,6 @@ test_data_path=
 
 [model]
 label_binarizer_path=models/label_binarizer.pkl
-model_path=models/cnn-2020.04.3
+# model_path=models/cnn-2020.04.3
```
aCampello, line 19: Don't get why you comment this?

configs/2021.03.0.ini
```diff
@@ -0,0 +1,21 @@
+[DEFAULT]
+version=2021.02.0
```
aCampello, line 2: Version number different from the `.ini` name above?

grants_tagger/train.py
```diff
@@ -17,6 +17,11 @@
 from grants_tagger.models import create_model
 from grants_tagger.utils import load_train_test_data, yield_tags
 
+from tensorflow.random import set_seed
+
+# TODO: Remove when WellcomeML implements setting random_seed inside models
+# replace with param in configs then
+set_seed(41)
```
aCampello, line 24: Not 42? :P

aCampello (approved):
Interesting. You could also add an observation at #results.md to say that the model size is much smaller, although f1 decreased a bit.

Some other minor comments in the review too.

docs/results.md
```diff
@@ -47,6 +47,8 @@ Version   | Approach    | Micro f1 | Description
 2020.10.15 | tfidf-svm | 0.71 | same as 2020.5.2 but train on title+synopsis and test on title+synopsis+research_question
 2020.10.16 | tfidf+onehot_team | 0.71 | similar to 2020.5.2 but we add team as one hot variable to the features. as data get shuffled differently it seems as the performance stays the same but it actually increases 1% compared to removing the team variable
 2020.10.17 | tfidf+onehot_scheme | 0.68 | similar to 2020.10.16 but this time we add scheme as one hot variable.
+2021.03.0 | cnn | 0.60 | same as 2020.04.3 but with transformers tokenizer for encoder
+2021.03.1 | tfidf-transformers-svm | 0.68 | same as 2020.05.2 but tokenizer is TransformersTokenizer
```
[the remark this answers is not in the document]
nsorros: that was always the case. compare it with `2020.04.3` for a fair comparison.

configs/2020.04.3.ini
nsorros: Good catch, I need to change some of the configs now that I refactored the code as some non productionised models are saveable because they are neither pickable nor contain a save function which is fine since we only care about reproducing the result

configs/2021.03.0.ini
nsorros: good catch, I think I fixed it in a later commit but will check again

grants_tagger/train.py
nsorros: haha, well 42 performed very bad :(

configs/2021.03.0.ini
nsorros: nope I did not, will fix 😬

configs/2020.04.3.ini
nsorros: I am planning to do any changes necessary as part of the DVC transition which will also further improve the reproducibility

configs/2021.03.0.ini
nsorros: 6fa58ad

nsorros:
> Interesting. You could also add an observation at #results.md to say that the model size is much smaller, although f1 decreased a bit.
> 
> Some other minor comments in the review too.

btw in this case it does not matter very much that the model size is smaller as models for science tags are small anyway

PR #64, opened 2021-04-12
Add dvc

nsorros:
I am introducing DVC to the project. To do that I had to
* run dvc init that creates ~600 lines of .dvc files which you can ignore (half the PR)
* added data files to DVC and `.dvc` files for data to GitHub to track the version of the data
* create params.yaml to record the parameters that the pipelines will use. this is the equivalent to our configs.
* created pipelines for production models so currently tfidf-svm and scibert for science ensemble and mesh-cnn. in the near future I might replace the two pipelines that create the infividial components of science ensemble with a fit from science ensemble
* created a results folder and recorded results for the science runs. will also do for mesh before I close the PR, training for mesh takes a bit of time.

In regards to `params.yaml` and configs, I wanted to maintain the configs for backward compatibility and treat them as experiments for now. This could possibly be replaced by dvc experiments or not in the future. As such I am treating the config as always taking precedence in regards to setting params. I expect to use configs for running experiments and dvc repro and params.yaml for recording production pipelines. If I stick with that flow of a dual config, params.yaml, I might create a script that transforms a config to a params.yaml so that I can transition from an experiment to production easily.


nsorros:
[Transition to DVC](https://trello.com/c/bBsN2Zct/1192-transition-to-dvc)

grants_tagger/__main__.py
```diff
@@ -72,33 +103,65 @@ def train(
...
         output_path = cfg["preprocess"]["output"]
         mesh_metadata_path = cfg["filter_disease_codes"]["mesh_descriptions_file"]
-
-    if os.path.exists(output_path):
+        filter_tags = cfg["filter_disease_codes"].get("filter_tags")
+        test_split = cfg["preprocess"].get("test_split")
+
+    # TODO: Refactor with preprocess_mesh
+    if test_split:
+        data_dir, data_name = os.path.split(output_data_path)
```
lizgzil, line 133: should output_data_path be output_path?

grants_tagger/evaluate_model.py
```diff
@@ -2,17 +2,18 @@
 Combine pretrained models and evaluate performance
 """
 import pickle
+import json
 
 from sklearn.model_selection import train_test_split
-from sklearn.metrics import f1_score, classification_report, precision_score, recall_score
+from sklearn.metrics import classification_report, precision_recall_fscore_support
```
lizgzil, line 8: oo handy, didn't know about this

grants_tagger/__main__.py
nsorros: good catch, do not know how you find them and not sure how it passed the tests. aaa I know, I am not testing the cli very thoroughly, will add a test and fix

grants_tagger/evaluate_model.py
nsorros: sklearn is a bag of little treasures

grants_tagger/__main__.py
nsorros: Fixes here 347b64b but not adding a test now as I need some fake data for the xml which is a bit more work, will do in the future

lizgzil (approved):
it looks good to me

nsorros:
I will leave this open and merge on Monday in case @aCampello wants to take a look and suggest anything.

dvc.yaml
```diff
@@ -0,0 +1,132 @@
...
+    - preprocess_wellcome_science.text_cols
+    - preprocess_wellcome_science.meta_cols
+    outs:
+    - data/processed/science_grants_tagged_title_synopsis.jsonl
+  train_tfidf_svm:
+    cmd: grants_tagger train data/processed/science_grants_tagged_title_synopsis.jsonl
+      models/label_binarizer-2020.05.2.pkl models/tfidf-svm-2020.05.2.pkl --approach tfidf-svm
+    deps:
+    - data/processed/science_grants_tagged_title_synopsis.jsonl
+    - grants_tagger/train.py
```
aCampello, line 18: I'm very late to this game, but just noting that if you change any of the upstream files that this depend on (for example `utils` and `.yield_data`), dvc won't know it unless you explicitly put it is as a dependency.

params.yaml
```diff
@@ -0,0 +1,33 @@
+preprocess_wellcome_science:
+  text_cols: "Title,Synopsis"
+  meta_cols: "Grant_ID,Title"
+train:
+  tfidf-svm:
+    tfidf:
```
aCampello, line 6: As you mentioned, this is a bit redundant with our config files. Another way would be for the `params.yaml` to be minimal, just a path to the newest config file. Which one is the best I don't know...

aCampello (commented):
I know I am late to the game, but wanted to read this PR anyway. I like using DVC for productionised runs of the models. I'm not convinced that DVC experiments will help with my flow, since I like to use notebooks for quick experimentation.

params.yaml
aCampello: e.g.:

```
train:
   model_version:
   config_file_path:
````

then your train function would just look at parameters and trigger opening the config file. The dvc "diff" will be effectively a diff between model versions, but to actually see the different parameters you would have to go into the config file.

dvc.yaml
nsorros: This is a true, I think there is a fine balance to how much you add as dependencies so that you do not have to rerun repro every time but adding the utils is sensible.

params.yaml
nsorros: this way you lose all the experiments feature where you can see the different params but at the same time we do not use that feature. I am leaning more on the camp that we can created params.yaml from the config so essentially convert one to the other

nsorros:
> I know I am late to the game, but wanted to read this PR anyway. I like using DVC for productionised runs of the models. I'm not convinced that DVC experiments will help with my flow, since I like to use notebooks for quick experimentation.

Fair enough, how we use it is up to us. I am not fully convinced of experiments either. I think the most important feature that DVC brings in is the fact that you define the DAG of operations and you keep track of whether you changes have affected your pipelines.

PR #70, opened 2021-06-07
Split data sync to science and mesh

nsorros:
As I was not able to find a way to use `dvc --glob` to selectively pull and push, I am adding a way for the user to control which data to sync as a separate make command. This way the user can choose to sync the mesh project or not.

`dvc pull` pulls all the data but since `dvc` is meant for production, that is not necessarily bad. A user can safely experiment with the make commands.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,7 @@ models/**
 **/*.xml
 **/*.csv
 **/*.json
+**/*.txt
 !results/*.json
 
 # Hidden files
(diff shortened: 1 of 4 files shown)


nsorros:
[Fix sync data and models and DVC for big files and update README. Allow user to download all data if they want](https://trello.com/c/cWadEll7/1241-fix-sync-data-and-models-and-dvc-for-big-files-and-update-readme-allow-user-to-download-all-data-if-they-want)

nsorros:
Merging this as some changes are required to move on @aCampello. Happy to address comment in future PR.

PR #72, opened 2021-06-10
Add releases to github through tags and upload to PyPi through twine

nsorros:
The idea is to expand releasing grants tagger through `make deploy` (a better name is probable release) to create a release in Github and PyPi.

To release in Github we use GitHub tags and to release in PyPi we use `twine`


nsorros:
[Create and document releases](https://trello.com/c/ExhFif4j/1189-create-and-document-releases)

nsorros:
Btw the `username` and `password` I have for `PyPi` seem to not be correct. Can you send me the ones you use @aCampello? I assume we changed them at some point and I did not update them

Makefile
```diff
@@ -56,6 +56,9 @@ build: ## Create wheel distribution
 .PHONY: deploy
 deploy: ## Deploy wheel to public s3 bucket
 	aws s3 cp --recursive --exclude "*" --include "*.whl" --acl public-read dist/ s3://$(PUBLIC_PROJECT_BUCKET)
+	git tag v$(shell python setup.py --version)
```
aCampello, line 59: This is great, I really like tying the release to a tag. This way we version control the data too because of dvc, so we can have the cake and it eat it too.

CHANGELOG.md
```diff
@@ -0,0 +1,12 @@
+## v0.1.x
```
aCampello, line 1: why x?

grants_tagger/__init__.py
```diff
@@ -0,0 +1 @@
+__version__ = "0.1.3"
```
aCampello, line 1: 0.1.x above and 0.1.3 here?

aCampello (approved):
Great addition/great way to deploy. Since dvc controls data version through the `dvc.lock`, the release tag will also be tied to the latest data used.

A small comment on 0.1.x above, but otherwise very good idea.

CHANGELOG.md
nsorros: I was meaning two use version 0.1.x for experimentation with releases and start being more strict with 0.2

grants_tagger/__init__.py
nsorros: yes by x I mean any number should I use * or something else?

Makefile
nsorros: To be honest I am new to releases and tags. What is the alternative? To tag manually when you remember?

aCampello: No, I think this is best practice. What I meant was that in your case there is a bonus because data is also versioned through GitHub, so a tag will also point to a specific version of the data.

Potentially we can think of versioning models too through tags, like spacy does: https://github.com/explosion/spacy-models/releases.

This way a release tag encompasses everything, model, code and (through dvc.lock) data.

nsorros: Well dvc also versions models through `dvc repro` so if you go to a previous commit the model is rolled to whatever version was produced then

aCampello: Ah, yeah, that's right, even better.

PR #73, opened 2021-06-17
Rewrite tag grants to be more reusable

nsorros:
The intention of this PR it to make `grants_tagger tag` more general for potential users of the project but also so that it can be used in production incase @jdu this is a good idea. As such I have followed mainly what we have done in the production task and introduced some variables for the user to be able to define what is the field for id, text and what custom value might be written for null which in the case of our data warehouse is `No Data Entered`


grants_tagger/__main__.py
```diff
@@ -352,10 +352,16 @@ def tag(
         tagged_grants_path: Path = typer.Argument(..., help="path to output csv"),
         model_path: Path = typer.Argument(..., help="path to model"),
         label_binarizer_path: Path = typer.Argument(..., help="label binarizer for Y"),
-        threshold: float = typer.Option(0.5, help="threshold upon which to assign tag")):
+        approach: str = typer.Argument(..., help="approach used to train the model"),
+        threshold: float = typer.Option(0.5, help="threshold upon which to assign tag"),
+        grant_id_field: str = typer.Option("grant_id", help="field name for grant id"),
+        grant_text_fields: str = typer.Option("title,synopsis", help="comma separated text fields to be used for tagging"),
+        text_null_value: str = typer.Option("No Data Entered", help="value indicating null in text fields")):
```
aCampello, line 359: It would probably need a bit of re-factoring, but there is also a lot of junk synopses that @lizgzil and I collected over time to clean before training models: 

https://github.com/wellcometrust/datalabs/blob/master/modelling/grants/cleaning/common.json

Might be interesting to try and add to your pipeline at some point.

grants_tagger/tag_grants.py
```diff
@@ -8,63 +8,51 @@
...
+
+            for grant, tags in zip(grants, grants_tags):
+                for tag, prob in tags.items():
+                    csv_writer.writerow({
+                        'Grant id': grant[grant_id_field],
+                        'Tag': tag,
+                        'Prob': prob
+                    })
+
+            tagged_grants_tf.flush()
```
aCampello, line 58: do you need this? is it something to do with the way csv_writer handles I/O?

aCampello (approved):
The PR looks good as far as I am concerned. 

This probably means that https://github.com/wellcometrust/data-pipelines can be simplified a bit, but that's more a conversation with @jdu

grants_tagger/tag_grants.py
nsorros: tbh I copied and pasted from https://github.com/wellcometrust/data-pipelines/blob/739534868294dd0047934e859800a800f5c5c3bd/pipelines/mesh/task_tagger.py#L81

maybe @jdu can enlighten us if this is needed

jdu: Basically, when you write to a file, in this case calling `writerow` it doesn't actually write to the file immediately, it writes to a buffer. That buffer gets "flushed" to the file periodically. What we're doing here by explicitly calling `flush` is making sure that every time we start another batch of 128 grants, we flush the results to disk. 

It reduces memory overhead so you're not carrying around an ever-growing buffer for a long period of time. 

Usually when you use something like

```
with open("somefile.csv", "wb") as f:
    f.writerow([])
```

python automatically flushes the buffer to disk once you leave the `with` block, but with a longer running process like grant tagger, mesh tagger, etc... where we want to flush that buffer manually periodically to keep our memory usage down.

jdu:
I don't know if this is necessary, mainly because it's prescriptive, for example, our grants don't live in a file on the file system, they live in an RDBMS and I need to get them through a database driver, or I might be storing them in a parquet, jsonl, yaml, etc.... and I can't parallelize the work amongst multiple workers using this without first dumping to a file, then splitting the file and passing each worker a specific file versus... just getting it direct from source on some partitioning scheme, I also want to write back to the database, not to a file, so now I have to have the additional step of reading in the output file to move it into it's final destination instead of just writing to the destination(s) I need. 

There needs to be a separation of concern here. There's no need for this library to understand CSVs or other file formats. The code that's here in this PR would probably make more sense under an `/examples` folder in the root of the repo to show people how to use the library but I don't think it needs to dictate how it's interacted with in this way.

I would much rather see a well-defined, clean and straighforward programmatic interface I can import for integrating the library than the library itself dictating how it's used. If I need to change the source, or have additional properties travel with the outputs from the classifications, I don't want to have to come back and rewrite the package itself, I just want to make the changes in my implementation code elsewhere.

```
from grant_tagger import Predictor

my_predictor = Predictor(...options...)
result = my_predictor.predict(content_to_tag_on, extra_params={"grant_id":  my_grant_id})
```

or 
```
from grant_tagger import Predictor

my_predictor = Predictor(...options...)

my_grants = [(text_string, dict(grant_id="my_grant_id"))]
results = my_predictor.batch_predict_tags(my_grants)
```

This allows me to be more flexible about the data that is being fed into the prediction method, it's up to the person leveraging the library to ensure that the data being passed in is sufficiently clean and correct. The library code itself doesn't need to know anything about the source structure of the data it is accepting, it only needs the data itself, and some additional params that should be returned along with the result (i.e. grant_id, or any additional params or identifiers that the user of the library needs to be associated with the result).

I think if the intent is to allow someone to quickly get a prediction on some content, you're better off moving this to an example in the repo like I said above as this library is more apt to be used wrapped inside of something else like API, local script, batch process, lambda function, etc...

grants_tagger/tag_grants.py
nsorros: thanks for the explanation, makes sense

aCampello: Nice! Thanks, I didn't know that about `writerow`!

nsorros:
> I don't know if this is necessary, mainly because it's prescriptive, for example, our grants don't live in a file on the file system, they live in an RDBMS and I need to get them through a database driver, or I might be storing them in a parquet, jsonl, yaml, etc.... and I can't parallelize the work amongst multiple workers using this without first dumping to a file, then splitting the file and passing each worker a specific file versus... just getting it direct from source on some partitioning scheme, I also want to write back to the database, not to a file, so now I have to have the additional step of reading in the output file to move it into it's final destination instead of just writing to the destination(s) I need.
> 
> There needs to be a separation of concern here. There's no need for this library to understand CSVs or other file formats. The code that's here in this PR would probably make more sense under an `/examples` folder in the root of the repo to show people how to use the library but I don't think it needs to dictate how it's interacted with in this way.
> 
> I would much rather see a well-defined, clean and straighforward programmatic interface I can import for integrating the library than the library itself dictating how it's used. If I need to change the source, or have additional properties travel with the outputs from the classifications, I don't want to have to come back and rewrite the package itself, I just want to make the changes in my implementation code elsewhere.
> 
> ```
> from grant_tagger import Predictor
> 
> my_predictor = Predictor(...options...)
> result = my_predictor.predict(content_to_tag_on, extra_params={"grant_id":  my_grant_id})
> ```
> 
> or
> 
> ```
> from grant_tagger import Predictor
> 
> my_predictor = Predictor(...options...)
> 
> my_grants = [(text_string, dict(grant_id="my_grant_id"))]
> results = my_predictor.batch_predict_tags(my_grants)
> ```
> 
> This allows me to be more flexible about the data that is being fed into the prediction method, it's up to the person leveraging the library to ensure that the data being passed in is sufficiently clean and correct. The library code itself doesn't need to know anything about the source structure of the data it is accepting, it only needs the data itself, and some additional params that should be returned along with the result (i.e. grant_id, or any additional params or identifiers that the user of the library needs to be associated with the result).
> 
> I think if the intent is to allow someone to quickly get a prediction on some content, you're better off moving this to an example in the repo like I said above as this library is more apt to be used wrapped inside of something else like API, local script, batch process, lambda function, etc...

This is interesting. In regards to predict api there is one `from grants_tagger.predict import predict_tags` which receives a model path, data_path etc. This is the one we use in tagging grants.

The purpose of tag was to emulate what were doing in production so that we can run it before we productions and give some early results to Kirstin. You might argue we might as well do that during deployment as part of qa but anyway that was the thinking. I was also thinking that maybe we can use this as is in the workflow so that we can maintain it here and test it here but the separation of concerns argument makes sense.

I wonder whether the argument to keep it as a means to harmonise what we do in production and here so that we can run locally and give to a stakeholder prior to deploying.

jdu:
Yeah after reading through this a second time I still think this probably shouldn't be directly integrated into the package. The entrypoint to grant_tagger should be relatively agnostic about where and what data comes at it really (source type, naming conventions,e tc...). If we're prescriptive about the file type, column names, etc... then we limit or needlessly complicate the adoption of the library for other people. What if my data is in a parquet file, or JSON(L), SQLite, postgres, etc.... Instead of trying to build conveniences for all these different use-cases, just create a nice generic programmatic API that allows them to feed data in as they see fit. 

Why do we need it to be a CLI tool? All the internal uses involve wrapping it in a larger python application or script which is perfectly fine. If I change my pipeline I don't want to have to refactor this interface, or add new interfaces to the package, I just want to update my local implementation around this package.

In the case of needing to be able to run it for validating correct-ness, etc... that doesn't necessarily need to be a part of the library itself, I think that's better served as something secondary in the repo that imports the package. But I think we should stray away from people thinking the package only accepts a CSV file (even though it has a programmatic API), and have these specific column names, and that we're going to make changes to the input data before it's predicted on DIRECTLY in the library itself as that limits it's use or complicates its use in different contexts and might result in people asking you to start writing code to support more formats, sources, outputs which isn't the point of this library. Reading and writing files and data is pretty trivial, low effort work in python so I don't think we need to provide that as a convenience.

What I as a user of this library would like to see, is a real world example of using the library in the repo, that I can potentially copy and paste and modify the bits I need to for my own use case.

That's just my two cents, but on the whole I wouldn't leverage this and would more than likely look at this as an example of how to implement and then roll my own implementation around it using whatever sources/targets I want to use bypassing this. So unless there are more uses of it I can't see, the justifiaction for it being directly embedded in the library is limited to testing & verification/validation which are QA processes?

nsorros:
> Yeah after reading through this a second time I still think this probably shouldn't be directly integrated into the package. The entrypoint to grant_tagger should be relatively agnostic about where and what data comes at it really (source type, naming conventions,e tc...). If we're prescriptive about the file type, column names, etc... then we limit or needlessly complicate the adoption of the library for other people. What if my data is in a parquet file, or JSON(L), SQLite, postgres, etc.... Instead of trying to build conveniences for all these different use-cases, just create a nice generic programmatic API that allows them to feed data in as they see fit.
> 
> Why do we need it to be a CLI tool? All the internal uses involve wrapping it in a larger python application or script which is perfectly fine. If I change my pipeline I don't want to have to refactor this interface, or add new interfaces to the package, I just want to update my local implementation around this package.
> 
> In the case of needing to be able to run it for validating correct-ness, etc... that doesn't necessarily need to be a part of the library itself, I think that's better served as something secondary in the repo that imports the package. But I think we should stray away from people thinking the package only accepts a CSV file (even though it has a programmatic API), and have these specific column names, and that we're going to make changes to the input data before it's predicted on DIRECTLY in the library itself as that limits it's use or complicates its use in different contexts and might result in people asking you to start writing code to support more formats, sources, outputs which isn't the point of this library. Reading and writing files and data is pretty trivial, low effort work in python so I don't think we need to provide that as a convenience.
> 
> What I as a user of this library would like to see, is a real world example of using the library in the repo, that I can potentially copy and paste and modify the bits I need to for my own use case.
> 
> That's just my two cents, but on the whole I wouldn't leverage this and would more than likely look at this as an example of how to implement and then roll my own implementation around it using whatever sources/targets I want to use bypassing this. So unless there are more uses of it I can't see, the justifiaction for it being directly embedded in the library is limited to testing & verification/validation which are QA processes?

Makes sense, I will remove it then and add it to scripts

PR #74, opened 2021-06-28
Add grants_tagger visualize

nsorros:
Adds `grants_tagger vizualize` command that runs a [streamlit](https://streamlit.io) app. It allows the user to select a model, threshold and whether to display probabilities and then paste any text to get tags

Screenshot 2021-06-28 at 2 06 19 PM


grants_tagger/__main__.py
```diff
@@ -362,6 +363,10 @@ def explain():
     # feature importance for models
     pass
 
+@app.command()
+def visualize():
+    st_app_path = os.path.join(os.path.dirname(__file__), "streamlit_visualize.py")
+    subprocess.Popen(["streamlit", "run", st_app_path])
```
aCampello, line 369: Wonder if there is a way to run Streamlit programatically, like Flasks' `import app; app.run()`

grants_tagger/streamlit_visualize.py
```diff
@@ -0,0 +1,48 @@
+import streamlit as st
+import pandas as pd
+
+from grants_tagger import predict_tags
+
+threshold = st.sidebar.slider("Threshold", min_value=0.0, max_value=1.0, value=0.5)
+text = st.text_area('Grant abstract', 'The cell is...', height=300)
+
+models = {
+    "disease_mesh_cnn-2021.03.1": {
```
aCampello, line 10: Nice model dispatcher, although in the future you can probably accept the model config file as parameter to your `grants_tagger visualize`.

grants_tagger/streamlit_visualize.py
```diff
@@ -0,0 +1,48 @@
...
+        "approach": "scibert"
+    }
+}
+
+model_option = st.sidebar.selectbox("Model", options=list(models.keys()))
+model = models[model_option]
+
+probabilities = st.sidebar.checkbox("Display probabilities")
+
+with st.spinner('Calculating tags...'):
```
aCampello, line 37: So easy.

aCampello (approved):
Code looks good, and from the video/pictures the visualisation looks quite good too! Streamlit syntax seems nice and clean.

Seems to have a conflict with requirements.txt to fix before merging.

grants_tagger/__main__.py
nsorros: I did not find a way, but will look again. Btw @jdu is this a good way if invoking a script? I remember hunter had mentioned that some python way was better than another

grants_tagger/streamlit_visualize.py
nsorros: <3 streamlit

grants_tagger/__main__.py
jdu: There's a few different ways of spawning a sub-process across the different python versions. This is probably fine for what is just a labs component.

PR #76, opened 2021-07-05
Add download EPMC script

### LINKED ISSUE #54: Implement download function to download training data from EPMC

nsorros:
This PR adds a `download` command to grants tagger so that the Mesh data can be downloaded from EPMC. The rationale is that BioASQ provides the data behind a registration whereas with EPMC it is freely available for all to download without the need to register.

I have copied and reimplemented the EPMC client as because of the volume of the data I needed it to write on the fly so that is more efficient and also handle partial downloads. If these are features we want in WellcomeML we can work towards bringing the two together in the future.

Fixes #54

### DIFF
--- a/grants_tagger/download_epmc.py
+++ b/grants_tagger/download_epmc.py
@@ -26,7 +26,7 @@ def _get_response(session, params):
 def yield_results(session, params):
     response_json = _get_response(session, params) 
 
-    results = response_json["resultList"]["result"] 
+    results = response_json["resultList"]["result"]
     while results:
         for result in results:
             yield result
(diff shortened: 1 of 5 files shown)


grants_tagger/download_epmc.py
```diff
@@ -0,0 +1,71 @@
...
+        tmp_month_path = f"{month_path}.tmp"
+        with open(tmp_month_path, "w") as f:
+            params = {
+                "query": f"(FIRST_PDATE:[{year}-{month}-01 TO {year}-{month}-31])",
+                "format": "json",
+                "resultType": "core",
+                "pageSize": 100
+            }
+            hit_count = get_hit_count(session, params)
+            for result in tqdm(yield_results(session, params), total=hit_count, desc=f"Year {year} Month {month}"):
```
aCampello, line 67: Nice loading bar

grants_tagger/download_epmc.py
```diff
@@ -0,0 +1,71 @@
...
+            print(f"Skipping because {month_path} exists. Delete if you want to redownload")
+            continue
+
+        tmp_month_path = f"{month_path}.tmp"
+        with open(tmp_month_path, "w") as f:
+            params = {
+                "query": f"(FIRST_PDATE:[{year}-{month}-01 TO {year}-{month}-31])",
+                "format": "json",
+                "resultType": "core",
+                "pageSize": 100
```
aCampello, line 64: Could be a parameter.

aCampello (approved):
Very nice. Impressed with how comprehensive the tests are ⭐ !

Definitely +1 to moving the partial monthly data save features to WellcomeML in the future If you want.

tests/test_download_epmc.py
```diff
@@ -0,0 +1,61 @@
...
+
+from grants_tagger.download_epmc import download_epmc
+
+
+@pytest.fixture
+def download_path(tmp_path):
+    return tmp_path
+
+
+def create_month_data(month_path):
```
aCampello, line 15: Is this going to actually create data and delete it after the test?

aCampello: Or is it automatically deleted because of the fixture?

nsorros: automatically deleted because of the fixture

PR #77, opened 2021-07-07
Makefile.streamlit

jdu:
Adds a new Makefile.streamlit for building and deploying the streamlit container to ECR for deployment to the cluster.

### DIFF
--- a/Makefile.streamlit
+++ b/Makefile.streamlit
@@ -1,0 +1,20 @@
+AWS_ACCOUNT_ID := 160358319781
+ECR_IMAGE := 160358319781.dkr.ecr.eu-west-1.amazonaws.com/org.wellcome/labs
+TAG := streamlit-grants-tagger-latest
+
+.PHONY: aws-docker-login
+aws-docker-login:
+	aws ecr get-login-password --region eu-west-1 | docker login --username AWS --password-stdin $(AWS_ACCOUNT_ID).dkr.ecr.eu-west-1.amazonaws.com
+
+.PHONY: build-streamlit-docker
+build-streamlit-docker: ## Builds Docker with streamlit and models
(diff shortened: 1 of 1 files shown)


jdu:
This can be used by passing the path to the Makefile, I made a separate Makefile as it's only concerned with building the docker container for streamlit and not with anything else really. 

You can run the commands in it by passing the `-f` arg to `make`

i.e.

```
make -f Makefile.streamlit docker-push
```

Makefile.streamlit
```diff
@@ -0,0 +1,20 @@
+AWS_ACCOUNT_ID := 160358319781
```
nsorros, line 1: note that the repo is public

nsorros (approved):
LGTM 👍 Note the some env variables you are exposing. In fact the user is instructed to have some of them in their env https://github.com/wellcometrust/grants_tagger/blob/master/README.md#-env-variables

jdu:
Ah right, I forgot this repo is public, it might make sense then to move this Makefile elsewhere as it's specific to our infrastructure.

nsorros:
What should we do with this @jdu?

Makefile.streamlit
jdu, line 1:
```suggestion
AWS_ACCOUNT_ID := ${AWS_ACCOUNT_ID}
```

Makefile.streamlit
```diff
@@ -0,0 +1,20 @@
+AWS_ACCOUNT_ID := ${AWS_ACCOUNT_ID}
+ECR_IMAGE := 160358319781.dkr.ecr.eu-west-1.amazonaws.com/org.wellcome/labs
```
jdu, line 2:
```suggestion
ECR_IMAGE := ${AWS_ACCOUNT_ID}.dkr.ecr.eu-west-1.amazonaws.com/org.wellcome/labs
```

PR #78, opened 2021-07-08
Download model command

### LINKED ISSUE #75: Make mesh or all models public

nsorros:
Adds functionality to download publicly available models at the moment only `disease_mesh`. At the same time uploads those models when we do a new release.

Fixes #75

### DIFF
--- a/grants_tagger/download_model.py
+++ b/grants_tagger/download_model.py
@@ -7,7 +7,7 @@ import requests
 MODELS = {
     'disease_mesh': {
         'url': 'https://github.com/wellcometrust/grants_tagger/releases/download/v0.1.3/disease_mesh_cnn-2021.03.1.tar.gz',
-        'path': 'models/disease_mesh_cnn-2021.03.1.tar.gz'
+        'path': 'disease_mesh_cnn-2021.03.1.tar.gz'
     }
 }
 
@@ -29,7 +29,7 @@ def download_model(model_name):
     if model_name in MODELS:
         url = MODELS[model_name]['url']
         path = MODELS[model_name]['path']
-        models_path = os.path.join(os.path.dirname(__file__), '../models/')
+        models_path = os.path.join(os.path.dirname(__file__), '../')
         
         download_tar(url, path)
         untar(path, models_path)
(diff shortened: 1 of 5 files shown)


nsorros:
Have a look at this PR @aCampello when you have a moment

grants_tagger/download_model.py
```diff
@@ -0,0 +1,40 @@
+import tarfile
+import os
+
+from tqdm import tqdm
+import requests
+
+MODELS = {
+    'disease_mesh': {
+        'url': 'https://github.com/wellcometrust/grants_tagger/releases/download/v0.1.3/disease_mesh_cnn-2021.03.1.tar.gz',
+        'path': 'disease_mesh_cnn-2021.03.1.tar.gz'
```
aCampello, line 10: Are you going to update this line every time there is a new model? That might be ok in the short run as we don't have many models, but in the long run it'd be ace if you had a `get_latest_version` function so that the dispatcher only has a "base" path `disease_mesh_cnn`. I think spacy has it for free, because their models are also released as pypi packages, but you could emulate that too.

So the user can either `grants_tagger download disease_mesh` and that will give you latest or `grants_tagger download disease_mesh_cnn-2021.03.1.tar.gz`

grants_tagger/download_model.py
```diff
@@ -0,0 +1,40 @@
...
+}
+
+
+def download_tar(url, path):
+    response = requests.get(url, stream=True)
+    total_length = int(response.headers.get('content-length'))
+    
+    #iif response.status_code == 200:
+    with open(path, "wb") as f:
+        for chunk in tqdm(response.iter_content(chunk_size=1024), total=int(total_length / 1024), unit='KB'):
```
aCampello, line 21: very nice.

aCampello (approved):
The download command looks good! I have a small suggestion about getting the latest model if you want to implement now, or add it to the future.

grants_tagger/download_model.py
nsorros: I was actually thinking that the user will always download the model compatible with the version released. So when a new version is released in theory models would need to be retrained and this updated.

PR #90, opened 2021-09-29
Add XLinear for MeSH

nsorros:
Todo
* Add config
* Add to DVC
* Run DVC and update docs


aCampello:
Yeah, running this again, I think it needs a `try catch` because even `grants_tagger --help` wouldn't work due to pecos.

PR #137, opened 2021-11-05
Add autoblack github action

nsorros:
## Description

Add github action to run autoblack every Friday
and open a PR, inspired from SpaCy
https://github.com/explosion/spaCy/blob/master/.github/workflows/autoblack.yml

## Checklist

- [ ] Added tests
- [ ] Updated README
- [ ] DVC up to date

### DIFF
--- a/.github/workflows/autoblack.yml
+++ b/.github/workflows/autoblack.yml
@@ -15,9 +15,13 @@ jobs:
         with:
           ref: ${{ github.head_ref }}
       - uses: actions/setup-python@v1
-      - run: pip install black
+      - run: |
+          pip install black
+          pip install dvc
       - name: Auto format code if needed
-        run: black grants_tagger
+        run: |
+          black grants_tagger
+          dvc commit
       - name: Check for modified files
         id: git-check
         run: echo ::set-output name=modified::$(if git diff-index --quiet HEAD --; then echo "false"; else echo "true"; fi)
--- a/.github/workflows/autoblack.yml
+++ b/.github/workflows/autoblack.yml
@@ -22,7 +22,7 @@ jobs:
       - name: Auto format code if needed
         run: |
           black grants_tagger
-          dvc commit
+          dvc commit -f
       - name: Check for modified files
         id: git-check
         run: echo ::set-output name=modified::$(if git diff-index --quiet HEAD --; then echo "false"; else echo "true"; fi)
--- a/.github/workflows/autoblack.yml
+++ b/.github/workflows/autoblack.yml
@@ -3,6 +3,7 @@
 
 name: autoblack
 on:
+  pull_request:
   workflow_dispatch:
   schedule:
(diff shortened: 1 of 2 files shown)


.github/workflows/autoblack.yml
```diff
@@ -0,0 +1,34 @@
...
+
+jobs:
+  autoblack:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v1
+        with:
+          ref: ${{ github.head_ref }}
+      - uses: actions/setup-python@v1
+      - run: pip install black
```
aCampello, line 18: pip install dvc

.github/workflows/autoblack.yml
```diff
@@ -0,0 +1,34 @@
...
+      - name: Auto format code if needed
+        run: black grants_tagger
```
aCampello, line 20: run dvc commit

aCampello (approved):
Happy to approve to, but can we add `dvc commit`  so it forces the dvc pipeline to accept the changes?

nsorros:
To be honest I will close this and open a new PR for doing a black check which will allow the user to fix as they see fit. I tried doing `dvc commit` but you need access to s3 as far as I understand it. We can revisit how to make this work later.

ivyleavedtoadflax:
Yes @nsorros we will need a cml user to be able to push to s3. Have a look at the pipeline we use for oms-rasa-2, we basically do the exact same thing.

nsorros:
> Yes @nsorros we will need a cml user to be able to push to s3. Have a look at the pipeline we use for oms-rasa-2, we basically do the exact same thing. You need to get the AWS_SECRET_ACCESS_KEY and AWS_ACCESS_KEY_ID and add them as secrets to the repo on github. Then you need to make these available in the dvc commit stage in the `env` attribute.
> 
> In that last failed test, it looks like you would also need to do `dvc pull` or `dvc repro --pull` first to make sure all the pre-requisites are in place.

Yep I know, let's not share AWS creds in this project. I don't think CML is needed just the creds but I feel its much better to add checks for `flake8` and `black` and leave to user to `dvc commit` as they see fit. See #148. We can do it separately at Mantis where we already use CML.

aCampello:
We can share creds as GitHub secrets btw, nothing wrong with it - we've done for other projects.

nsorros:
> We can share creds as GitHub secrets btw, nothing wrong with it - we've done for other projects.

Ideally we would create fine grained creds that only have access to the bucket that dvc tries to use. I don't think its a good idea to have our creds in that have super access. Up to you and @jdu. Happy to reopen this.

@aCampello Note that the current flow has added two checks respectively for `black` and `flake8` so whoever opens a PR knows that there is a violation they need to fix. Additionally, they can resolve dvc issues with `dvc commit` as they sit fit while running those tools.

PR #143, opened 2021-11-09
Release 0.2.0 that includes XLinear

### LINKED ISSUE #95: Document which models/code version are in the production pipeline
Could be either in this repository or here: https://github.com/wellcometrust/data-pipelines

### LINKED ISSUE #110: Release XLinear and align model versioning with package
Once the branch that adds `XLinear` is merged we should create a new release. As part of the release we should remove the confusion that arises from having `CalVer` versions for models and `SemVer` for the package.

I see two options forward
* we either switch to `SemVer` when transitionining from experimentation to production. `CalVer` is still useful for experimenting to track timeline.
* or we can document which `calver` version is the one in production

Maybe this is a good time to start using DVC for picking up model versions.

### LINKED ISSUE #142: train command fails locally due to ConnectionError
This also affects `dvc repro` which uses the `train` command. The bug was introduced in #100 (@aCampello) because apparently `requests.get` throws a `ConnectionError` instead of returning with a status code locally.

nsorros:
## Description

We are releasing `0.2.0` that includes the `XLinear` model
which is trained in all MeSH.

Fixes #95 @aCampello to confirm
Fixes #110
Fixes #142
[Notion card](https://www.notion.so/wellcometrust/Release-XLinear-d2f877c9ab8142b7b506b3a4f283b0b3)

## Checklist

- [x] Added tests
- [x] Updated README
- [x] DVC up to date

## Release checklist

- [x] DVC repro up to date
- [x] Models synced in S3
- [x] Version updated
- [x] CHANGELOG.md updated
- [x] Model and package version aligned

To release:

* `make build`
* `make deploy`


dvc.yaml
```diff
@@ -1,3 +1,6 @@
+vars:
+  - version: 0.2.0
```
aCampello, line 2: Nice. Does it mean that the package version will be the same as the models version ?

grants_tagger/utils.py
```diff
@@ -104,7 +104,10 @@ def calc_performance_per_tag(Y_true, Y_pred, tags):
 def get_ec2_instance_type():
     """Utility function to get ec2 instance name, or empty string if not possible to get name"""
 
-    instance_type_request = requests.get('http://169.254.169.254/latest/meta-data/instance-type')
+    try:
```
aCampello, line 107: My bad, thanks for fixing this.

aCampello (approved):
FTW!

dvc.yaml
nsorros: Yes, I am looking into a way to pick up the version from the package as well

nsorros:
Newest `grants_tagger` released `0.2.0` https://pypi.org/project/grants-tagger/#history 🎉

nsorros:
⚠️ Note that model is 2.3GB which is larger than what is allowed from Github so we might need to point the `download command` to s3 and inform the user

nsorros:
Actually merging this as the `download model` command was not updated before pushing which means that it will download the previous model. Will fix with `0.2.1`

PR #147, opened 2021-11-10
Feature/fast tf idf

aCampello:
## Description

This PR adds the possibility of using the parallel version of the tf-idf vectoriser and makes MeshXLinear sklearn compatible.

With this, I believe it will unlock us some experiment like increasing max_features and ngram #149 .

It also adds a small script to notify slack after training finished.

[Notion card](https://www.notion.so/wellcometrust/dc03328f311d4676b497ec7cd6f51042?v=e1a6758d760d4689b24356826b2d3541&p=70909bf9f59749d3a68cae865dc120b2)

## Checklist

- [x] Added tests (Modified tests so it runs)
- [x] Updated README
- [ ] DVC up to date: 🚫 

## Release checklist

- [ ] DVC repro up to date
- [ ] Models synced in S3
- [ ] Version updated
- [ ] CHANGELOG.md updated

To release:

* `make build`
* `make deploy`


.envrc
```diff
@@ -0,0 +1,9 @@
+export PROJECTS_BUCKET=datalabs-data
+export PROJECT_NAME=grants_tagger
+
+export AWS_ACCOUNT_ID=160358319781
+export ECR_IMAGE=xxx
+export SAGEMAKER_ROLE=xxx
+
+export SLACK_HOOK=xxx
+export SLACK_USER=xxx
```
nsorros, line 9: did you mean to share this? better to populate the `.envrc.template` and let the user have their own `.envrc`. also note that if you populate the values git will detect a change if the file is git tracked.

grants_tagger/models/mesh_xlinear.py
```diff
@@ -25,70 +27,120 @@ def __init__(
...
+        else:
+            self.vectorizer_.save(model_path)
+
+        self.xlinear_model_.save(model_path)
 
     def load(self, model_path, is_predict_only=True):
         vectorizer_path = os.path.join(model_path, "vectorizer.pkl")
-        self.vectorizer = load_pickle(vectorizer_path)
-        self.model = XLinearModel.load(model_path, is_predict_only=is_predict_only)
+        if self.vectorizer_library == "sklearn":
```
nsorros, line 138: not that this is a bad strategy but what would be even better is to save the params in a `meta.json` and initialize from there so that you don't need to pass params when loading the model. otherwise a user might save a vectorizer and then try to load without knowing which vectorizer was used. the information should ideally be stored somewhere. `spacy` uses a `meta.json` as well. I have short of tried following that route but it is not complete yet in other models so its optional

nsorros (approved):
LGTM 🚀 I can't believe that a parallel tfidf was hiding under our nose. Have a think about whether params should be saved in the model_path somehow.

nsorros:
⚠️ Wait till the release goes out first #143 @aCampello

.envrc
aCampello: Oopsy. Fixed https://github.com/wellcometrust/grants_tagger/pull/147/commits/a435335ec02be91ca14630551f5e84063fc19325 and added a gitignore too

grants_tagger/models/mesh_xlinear.py
aCampello: Yep, meta route is definitely better. Let's add this to improvements!

nsorros:
@aCampello do you want me to take over and run `dvc repro`? You can also do btw before you merge. Or we can merge your other PR as well before running.

aCampello:
Maybe we can merge this, and then when I submit the optimisation one (which will be today), we can run dvc repro?

nsorros:
> Maybe we can merge this, and then when I submit the optimisation one (which will be today), we can run dvc repro?

Sounds good

nsorros:
🎉

PR #148, opened 2021-11-10
Format code with black and flake8

nsorros:
## Description

Runs `black grants_tagger` and fixes errors from `flake8 grants_tagger`. Also ignores certain error codes that are conflicting between black and flake8 or that will be addressed later.

## Checklist

- [ ] Added tests
- [ ] Updated README
- [ ] DVC up to date

### DIFF
--- a/.github/workflows/black.yml
+++ b/.github/workflows/black.yml
@@ -1,0 +1,12 @@
+name: Check black compliance
+on: pull_request
+
+jobs:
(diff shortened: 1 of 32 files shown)


aCampello (approved):
Looks good. I only looked at the workflow (black.yml), because it's impossible to review all files (we'll need to believe black and the tests 🙏 )

PR #151, opened 2021-11-12
Release 0.2.1

### LINKED ISSUE #152: Grants tagger broken due to dependencies from WellcomeML
As I was trying to test whether `grants_tagger download model mesh` works for release `0.2.1` in an isolated environment I stumbled upon a rabbit hole of problems related missing packages e.g.
`ImportError: To use this class/module you need to install wellcomeml with torch extras, e.g. pip install wellcomeml[torch]`

This is caused from the fact that we only install `wellcomeml[tensorflow]` but use a couple of modules from wellcomeml inside `create_model`. This is not detected as we run tests in the development environment where all dependencies are installed instead of an isolated environment like we have proposed in #141.

This affects #151 but it is also a good opportunity to move to latest release of WellcomeML #146.

I suggest we bump to wellcomeml `2.0`, then use direct imports and if the package is missing throw a warning to the user which can be disabled from setting the right log level. Additionally I suggest we switch to using tox to be able to catch these errors.

nsorros:
## Description

This release fixes `grants_tagger` package which was not working 
due to dependencies missing #152 and updates `grants_tagger download`
which was pointing to the disease_mesh model instead of the latest mesh.

As part of fixing the grants_tagger dependencies we upgrade to wellcomeml 2.0.0 🎉 

⚠️ To resolve future issue with updating version and not needing to rerun models
and since models are versioned through dvc anyway, we remove the version number
from the name and only include it in the packaged tarball. This means that we can package
models trained with the previous version if they are still compatible.

⚠️ We publish the models using our `datalabs-public` folder instead of Github releases
since the size is over 2GB which is the limit.

Fixes #152 
Notion card: https://www.notion.so/wellcometrust/Fix-broken-grants_tagger-package-14ba9244d3b643c1bd6e554c20f52f57

## Checklist

- [x] Added tests
- [x] Updated README
- [x] DVC up to date

## Release checklist

- [x] DVC repro up to date
- [x] Models synced in S3
- [x] Version updated
- [x] CHANGELOG.md updated
- [x] Model and package version aligned
- [x] Download model updated

To release:

* `make build`
* `make deploy`


nsorros:
@jdu this PR updates the link to point to the public repository as well as the download command.

Makefile
```diff
@@ -75,20 +81,29 @@ update-requirements: ## Updates requirement
...
 
 .PHONY: deploy
 deploy: ## Deploy wheel to public s3 bucket
 	aws s3 cp --recursive --exclude "*" --include "*.whl" --acl public-read dist/ s3://$(PUBLIC_PROJECT_BUCKET)
 	git tag v$(shell python setup.py --version)
 	git push --tags
 	$(VIRTUALENV)/bin/python -m twine upload --repository pypi dist/*
-	tar -c -z -v -f models/$(MESH_MODEL).tar.gz models/$(MESH_MODEL) models/$(MESH_LABEL_BINARIZER)
-	gh release upload v$(shell python setup.py --version) models/$(MESH_MODEL).tar.gz
+	aws s3 cp --acl public-read models/$(MESH_MODEL_PACKAGE) s3://datalabs-public/grants_tagger/models/$(MESH_MODEL_PACKAGE)
```
aCampello, line 98: I think it's ok to do it like this for now, and fir future release we'll find better ways

README.md
```diff
@@ -376,7 +376,7 @@ Commands:
 
 Available models:
 
-* disease_mesh
```
aCampello, line 379: disease mesh is also available no?

grants_tagger/models/create_model.py
```diff
@@ -33,20 +41,21 @@
...
-    KerasVectorizer,
-    SpacyClassifier,
-    BertVectorizer,
-    BertClassifier,
-    Doc2VecVectorizer,
-    Sent2VecVectorizer,
-    WellcomeVotingClassifier,
-    TransformersTokenizer,
-)
+from wellcomeml.ml.bilstm import BiLSTMClassifier
```
aCampello, line 44: nice!

aCampello (approved):
This is a huge PR :) Thanks! Let's release this so then it can be put into a pipeline and we can move on to the next thing.

aCampello:
Just noticed the pipeline is still broken - feel free to fix the test/comment it out and then merge.

Makefile
```diff
@@ -3,8 +3,14 @@
...
 PUBLIC_PROJECT_BUCKET := datalabs-public/$(PROJECT_NAME)
-MESH_MODEL := disease_mesh_cnn-2021.03.0
-MESH_LABEL_BINARIZER := disease_mesh_label_binarizer.pkl
+
+PACKAGE_VERSION := $(shell venv/bin/python -c "import grants_tagger;print(grants_tagger.__version__)")
+MESH_MODEL_PACKAGE := xlinear-$(PACKAGE_VERSION).tar.gz
+MESH_MODEL := xlinear/model/
+MESH_LABEL_BINARIZER := xlinear/label_binarizer.pkl
+SCIENCE_TFIDF_SVM_MODEL:= tfidf-svm.pkl
+SCIENCE_SCIBERT_MODEL := scibert
```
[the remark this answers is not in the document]
nsorros: yes and dropped the version from the name

README.md
nsorros: well I am replacing it with the entire mesh, we can make both available but we have created them through `dvc repro` so there is a minor chance there might be incompatible

PR #158, opened 2021-11-19
Feature/script to pull new grants

aCampello:
## Description

Towards: https://www.notion.so/wellcometrust/dc03328f311d4676b497ec7cd6f51042?v=e1a6758d760d4689b24356826b2d3541&p=b76ea39b19594ddbbc1c26182982d9d3

I added a script to update grants.csv file, but did not create a dvc.yaml stage for it, because it would block people that don't have access to `wellcometrust/datascience` (private repository) to run `dvc repro`

## Checklist

- [ ] Added tests
- [ ] Updated README
- [ ] DVC up to date

## Release checklist

- [ ] DVC repro up to date
- [ ] Models synced in S3
- [ ] Version updated
- [ ] CHANGELOG.md updated

To release:

* `make build`
* `make deploy`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -132,6 +132,12 @@ build-streamlit-docker: ## Builds Docker with streamlit and models
 	aws s3 cp s3://datalabs-data/grants_tagger/models/label_binarizer.pkl models/
 	docker build -t streamlitapp -f Dockerfile.streamlit .
 
+.PHONY: install-private-requirements
+install-private-requirements: ## Install the private datascience utils
+	pip install pyodbc psycopg2
+	pip install -e git+ssh://git@github.com/wellcometrust/datascience.git#egg=wellcome-datascience-common
+ 
+
 help: ## Show help message
 	@IFS=$$'\n' ; \
 	help_lines=(`fgrep -h "##" $(MAKEFILE_LIST) | fgrep -v fgrep | sed -e 's/\\$$//' | sed -e 's/##/:/'`); \
--- a/scripts/get_grants.py
+++ b/scripts/get_grants.py
@@ -8,7 +8,7 @@ try:
 except ImportError as e:
     raise ImportError(f"Error importing {e.name }. "
                       f"To use this script you need to install Wellcome's datascience internal "
-                      f"utils.")
+                      f"utils, for example with `make install-private-requirements`")
 
 forty_two = FortyTwo()
 df = pd.DataFrame(forty_two.get_grants())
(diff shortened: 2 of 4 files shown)


scripts/tag_grants_mesh_pipeline.sh
```diff
@@ -0,0 +1,7 @@
+python /pipelines/mesh/task_tagger.py \
```
nsorros, line 1: is this script supposed to be here?

scripts/get_grants.py
```diff
@@ -0,0 +1,25 @@
+#!/bin/bash
+import pandas as pd
+
+import os
+
+from datascience.warehouse.warehouse import FortyTwo
```
nsorros, line 6: datascience is not installed though?

aCampello: Yep, and it won't be because it's private. Maybe I'll add a `requirements-private` or something? What do you think?

scripts/tag_grants_mesh_pipeline.sh
aCampello: Lol, I committed this file by accident, ignore it, will delete.

scripts/get_grants.py
nsorros: or a make command that installs it for those that have access. it can be separate from `make virtualenv`. also a `try except` in the script to remind us to install it :)

scripts/tag_grants_mesh_pipeline.sh
aCampello: Done

scripts/get_grants.py
aCampello: Done https://github.com/wellcometrust/grants_tagger/pull/158/commits/ec4259e68434286256b624e5b6f3db8570bc3dd2

nsorros:
It is worth saying that users that do not have access to our AWS they cannot run `dvc repro` Or `dvc pull`. That said I think its fine if this remains separate. A user can download the latest data and add push the updated grants

Makefile
```diff
@@ -132,6 +132,12 @@ build-streamlit-docker: ## Builds Docker with streamlit and models
 	aws s3 cp s3://datalabs-data/grants_tagger/models/label_binarizer.pkl models/
 	docker build -t streamlitapp -f Dockerfile.streamlit .
 
+.PHONY: install-private-requirements
+install-private-requirements: ## Install the private datascience utils
+	pip install pyodbc psycopg2
```
nsorros, line 137: Is that not part of the `setup.py` Of datascience

nsorros (approved):

Makefile
aCampello: No - and I need to fix this. But for now we have requirements.txt only.

PR #162, opened 2021-11-29
update readme with pip information and basic example of prediction

### LINKED ISSUE #107: Update README with PyPi instructions and Quickstart guide
The following discrepancies need to be corrected in the README
* currently it states that the mode will be available in `PyPi` in the future. It is currently available in Pypi, we should include instructions to install and we should remove the mention to AWS as PyPi or Github releases is the preferred way.
* Bring your own model mentions that you need to update `models.py` but after `decouple-models` will be merged each file will be on its own file so the correct way would be to create a separate file and update `create_model.py` or if it is a simpler model simply to augment the `create_model`

Optional (if not implemented, add separate issues)
* Add quick start quide on how to use the "production models", in particular `MeshCNN` or `MeshXLinear`. It would be good to show how to download and use them for prediction as well as train them with your own data.

pdan93:
## Description
Update Readme with pip information.
Add basic example of downloading data and model and running a prediction.

Fixes #107 

## Checklist

- [ ] Added tests
- [x] Updated README
- [ ] DVC up to date


README.md
```diff
@@ -32,11 +32,17 @@ at the moment.
...
-manual installation. You need to git clone and pip install
-the code in your environment of choice.
+Easiest way to install is with pip.
+The project is available in the PyPI repository at https://pypi.org/project/grants-tagger/.
+
+```
+pip install grants-tagger
+```
+
+Alternatively you can clone this repository and build the project with
```
nsorros, line 42: This would not install the project, this would create the development environment for which we give instructions later. The manual installation would be to clone the project and run `pip install setup.py`

README.md
```diff
@@ -599,3 +604,23 @@ deploy               Deploy wheel to public s3 bucket
...
 help                 Show help message
 ```
+
+## 🔌 How to use production models
+
+We'll go through an example on how to download data and models, and use them for prediction and training using the cli commands.
+
+Data can be downloaded from here after you make an account: http://participants-area.bioasq.org/datasets/
+Download from the `Datasets for task a` section, the latest available txt format dataset. This will result in a zip file what you can unzip, preferably in the `data\raw` folder. 
+You can download pretrained model from the releases section of this repository. For the current example we'll a model trained with the `mesh-cnn` approach: https://github.com/wellcometrust/grants_tagger/releases/download/v0.1.3/disease_mesh_cnn-2021.03.1.tar.gz
```
nsorros, line 614: Note that the latest model could not be added to a release cause it was bigger than 2GB. So the standard way to download a model would be to use the CLI `grants_tagger download mesh`. That said assuming someone just wants to download the model we should add link that is updated in the README that either links to the release URL or s3, the link is the one in the code to download the model, this is just a manual way to get it.

pdan93: problem here is that it seems you need credentials to download from s3? I tried here to offer the public version.

nsorros: you should not need credentials so if you do, I need to fix that. its shared in a public bucket

pdan93: yeah, seems this worked:

pdan93: `grants_tagger download epmc-mesh models/`

pdan93: `grants_tagger download model mesh`
seems to have worked, so no need. I'll update the README

nsorros:
⚠️ Note that I switched to main for the base branch. I switched the base branch on the PR but for tests to work you need to cherry pick 469b116038bedae879f1ab7068832d198cf12f1d. Just run `git cherry-pick 469b116038bedae879f1ab7068832d198cf12f1d` if you aim to do more commits and want to run tests against latest main and not the soonoutdated master

nsorros:
Is this indeed ready for review @pdan93?

pdan93:
> Is this indeed ready for review @pdan93?

yes @nsorros

README.md
```diff
@@ -622,3 +627,28 @@ deploy               Deploy wheel to public s3 bucket
 clean                Clean hidden and compiled files
 help                 Show help message
 ```
+
+## 🔌 How to use production models
+
+We'll go through an example on how to download data and models, and use them for prediction and training using the cli commands.
+
+Data can be downloaded from here after you make an account: http://participants-area.bioasq.org/datasets/
```
nsorros, line 635: I think data are not needed for the basic user. My idea is that the basic user is not trying to use the data to train or anything but apply to their own data. They want to know how to download a model and use it to make predictions.

nsorros:
Actually to start with I like that the section on using the production model is small. I suggest we move it at the top and include it in the install and use section, not create a new section

nsorros:
Also dont forget to update the section Bring your own model

nsorros:
As I was re reading the README I noticed two additional discrepancies that we should fix
* in predict, it mentions a tag command that does not exist anymore, we should remove the mention
* in bting your own_data it mentions `__main__` which is not `cli`, we should update
@pdan93

pdan93:
I moved things to the install and use section, and cleaned it, and updated bring your own data. 
But the 'bring your own model' section, after looking at the code seems to be correct, what am I missing @nsorros ?

nsorros:
> I moved things to the install and use section, and cleaned it, and updated bring your own data.
> But the 'bring your own model' section, after looking at the code seems to be correct, what am I missing @nsorros ?

You are right, it looks good. Let me do a final review

nsorros (approved):
I am happy to merge as is but ideally there would be a Quickstart section that would demonstrate how someone can use the package or library in few commands to get predictions

aCampello:
Is this going to be merged?

PR #163, opened 2021-11-29
Add label binarizer creation to the preprocess step

pdan93:
## Description
Move the create_label_binarizer() in it's own file. Add it to the preprocess step as there was no way of using it through the CLI before.

## Checklist

- [ ] Added tests
- [ x ] Updated README
- [ ] DVC up to date

## Release checklist

- [ ] DVC repro up to date
- [ ] Models synced in S3
- [ ] Version updated
- [ ] CHANGELOG.md updated

To release:

* `make build`
* `make deploy`

### DIFF
--- a/grants_tagger/__main__.py
+++ b/grants_tagger/__main__.py
@@ -12,10 +12,12 @@ import time
 
 import typer
 
+
 logger = logging.getLogger(__name__)
 
 from grants_tagger.evaluate_mti import evaluate_mti
 from grants_tagger.evaluate_human import evaluate_human
+from grants_tagger.label_binarizer import create_label_binarizer
 from grants_tagger.pretrain import pretrain as pretrain_model
 from grants_tagger.evaluate_model import evaluate_model
 from grants_tagger.predict import predict_tags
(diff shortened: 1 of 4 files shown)


grants_tagger/label_binarizer.py
```diff
@@ -0,0 +1,36 @@
...
+    label_binarizer.fit(yield_tags(data_path))
+
+    with open(label_binarizer_path, "wb") as f:
+        f.write(pickle.dumps(label_binarizer))
+
+    return label_binarizer
+
+
+if __name__ == "__main__":
+    import argparse
```
nsorros, line 23: remove this

grants_tagger/__main__.py
```diff
@@ -272,6 +281,7 @@ def wellcome_science(
         print(f"{output_path} exists. Remove if you want to rerun.")
     else:
         preprocess(input_path, output_path, text_cols, meta_cols)
+        create_label_binarizer(output_path, label_binarizer_output_path)
```
nsorros, line 284: if we are to add this here, this should be in its own if then else that checks if the label_binarizer_path exists

grants_tagger/__main__.py
```diff
@@ -233,6 +239,9 @@ def wellcome_science(
     output_path: Optional[Path] = typer.Argument(
         None, help="path to JSONL output file that will be generated"
     ),
+    label_binarizer_output_path: Optional[Path] = typer.Argument(
```
nsorros, line 242: you don't like `label_binarizer_path`?

grants_tagger/__main__.py
nsorros: my initial though was that this could be added in the dvc command but I can see now that a separate cli command would need to be added to give the user the ability to split so maybe better like that.

nsorros (commented):
LGTM 👍 but let me think about is just a bit more before merging.

grants_tagger/__main__.py
pdan93: haha, yeah, I went with your `output_path` initially, but yep, changed it to be shorter

nsorros (approved):
Let's do it

pdan93:
First merge here, yeah!

PR #164, opened 2021-11-30
Add split train/test step in preprocessing

### LINKED ISSUE #115: Split train / test in separate script
Currently data are being split for `train/test` in `grants_tagger/utils.py` in `load_train_test_data`. This means that when we evaluate or want to retrieve the test set we need to use the same function and ensure we run it the same way which is not ideal and it lead to errors.

Instead we should split the data into a train / test set after preprocess and before train and use this split throughout.

Some considerations
* preprocess_mesh currently splits train and test in the absence of a separate script. the functionality needs to be removed from there
* evaluate_model and potentially other places use `load_train_test_data` to get a test set with the same split as in train. this should be removed together with the flag split data or the flag should be turned to false.
* `grants_tagger/utils.py` contains logic on how to split, this should be removed and moved to a separate file, in theory it should only contain logic how to load data in memory or via a generator

pdan93:
## Description

Add split_data.py script that splits into train/test. Use it also in preprocess commands.

Resolves #115 

## Checklist

- [x] Added tests
- [x] Updated README
- [x] DVC up to date


pdan93:
@nsorros still need to add tests, but check out if it's in line with what we talked

grants_tagger/__main__.py
```diff
@@ -287,6 +292,19 @@ def wellcome_science(
     else:
         create_label_binarizer(output_path, label_binarizer_path)
 
+    if test_split:
+        data_dir, data_name = os.path.split(output_path)
+        train_output_path = os.path.join(data_dir, "train_" + data_name)
+        test_output_path = os.path.join(data_dir, "test_" + data_name)
```
nsorros, line 298: I actually don't like that idea anymore, I think it should be explicitly passed as a param. So one param per train, test and test can be empty if the user does not want to split

grants_tagger/utils.py
```diff
@@ -1,14 +1,15 @@
...
 import pickle
 import json
 
 import requests
-from sklearn.model_selection import train_test_split
 from sklearn.metrics import f1_score
 import tensorflow as tf
 import pandas as pd
 import numpy as np
+from functools import partial
```
nsorros, line 10: do you use that?

nsorros: actually you just moved them it seems. is that black or you?

nsorros:
This is on a good path, just need to fix tests

grants_tagger/utils.py
pdan93: Honestly don't know, I think it was me cause at one point I had moved part of it in another file but then decided to put it back. 
WIll put them back in the same order they were.

grants_tagger/__main__.py
pdan93: @nsorros would you say in that case that we can get rid of output_path completely? and just have train_output_path and test_output_path?

nsorros: yep

grants_tagger/__main__.py
```diff
@@ -205,49 +212,59 @@ def bioasq_mesh(
...
         test_split = cfg["preprocess"].getfloat("test_split")
 
-    # TODO: Refactor with preprocess_mesh
-    if test_split:
-        data_dir, data_name = os.path.split(output_path)
-        train_output_path = os.path.join(data_dir, "train_" + data_name)
-        test_output_path = os.path.join(data_dir, "test_" + data_name)
+    if verify_if_paths_exist(
+        [
+            input_path,
```
nsorros, line 221: so if the input_path exists it returns?

grants_tagger/__main__.py
```diff
@@ -205,49 +212,59 @@ def bioasq_mesh(
...
+            test_output_path,
+        ]
+    ):
+        return
 
-        if os.path.exists(train_output_path) and os.path.exists(test_output_path):
-            print(
-                f"{train_output_path} and {test_output_path} exists. Remove them if you want to rerun."
-            )
+    temporary_output_path = train_output_path + ".tmp"
```
nsorros, line 229: why is that needed? to reduce space needed?

grants_tagger/__main__.py
```diff
@@ -205,49 +212,59 @@ def bioasq_mesh(
...
+    preprocess_mesh(input_path, temporary_output_path, mesh_tags_path=mesh_tags_path)
+    create_label_binarizer(temporary_output_path, label_binarizer_path)
+
+    if test_split:
```
nsorros, line 233: test_split is missing as a parameter to the CLI. also I think test_split should have a default value and only if test_output_path is None it should not split but not fixed on that. if you strongly feel otherwise keep as is.

grants_tagger/__main__.py
pdan93: hah! good catch!

grants_tagger/__main__.py
pdan93: yes, that's what I thought. Cause first you do the preprocessing and you need a file to output that to. And then you create the train/test from it.

grants_tagger/__main__.py
pdan93: hm, ok, I favor that also, 0.1 I suppose would be the default value @nsorros ?

nsorros:
Don't forget to link to the issue https://github.com/wellcometrust/grants_tagger/issues/115 to close it automatically when this merges

Also you can remove the release checklist in PR where this does not apply

grants_tagger/__main__.py
nsorros: 0.1 for science and 0.01 for bioasq.

grants_tagger/__main__.py
nsorros: and also this file is 20GB, good idea. if you were doing it to pass it in the label binarizer I would say that you can safely pass just the train there. but in the case of reducing size makes sense

pdan93:
@nsorros resolved comments, good to merge?

grants_tagger/utils.py
```diff
@@ -132,3 +134,20 @@ def load_pickle(obj_path):
...
+        f.write("\n")
+
+
+def verify_if_paths_exist(paths):
+    exist = 0
+    for path in paths:
+        if path and os.path.exists(path):
+            print(f"{path} exists. Remove if you want to rerun.")
+            exist += 1
+    if exist > 0:
```
nsorros, line 151: this is a bit confusing. the function is called verify if paths exist and returns false if they exist, i would anticipate that if paths exists this return True.

pdan93: fixed

pdan93:
@nsorros resolved the last small thing, good to go?

nsorros (approved):
LGTM 👍

PR #167, opened 2021-12-06
Feature/dvc dag to get grants and generate validation data

aCampello:
## Description

It adds a new independent dag to generate the validation data, useful in case we want to change anything:

```
+------------+              +-------------------------------------------+  
| get_grants |              | data/interim/mesh_pipeline_result.csv.dvc |  
+------------+***           +-------------------------------------------+  
                 ***                  ****                                 
                    ****          ****                                     
                        **      **                                         
           +----------------------------------+                            
           | generate_validation_data_xlinear |                            
           +----------------------------------+    
```

`dvc repro` won't be up-to-date.

## Checklist

- [ ] Added tests
- [ ] Updated README
- [ ] DVC up to date

## Release checklist

- [ ] DVC repro up to date
- [ ] Models synced in S3
- [ ] Version updated
- [ ] CHANGELOG.md updated

To release:

* `make build`
* `make deploy`


scripts/trigger_pipeline.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+
+# For this script to work, you need to have the repository data-pipelines pulled,
+# And place it in the root of that repository
+
+python pipelines/mesh/task_tagger.py \
```
nsorros, line 6: is that part of this PR?

nsorros:
`dvc.lock` seems to have changed is that because you run the step you added? You can get `dvc repro` up to date with your change as soon as #168 merges

data/interim/mesh_pipeline_result.csv.dvc
```diff
@@ -0,0 +1,4 @@
+outs:
+- md5: 10639bbe244b986919efc8f7866b98b4
+  size: 138862818
+  path: mesh_pipeline_result.csv
```
nsorros, line 4: did you add that manually?

nsorros:
Also do you intend for the notebook to be added?

nsorros:
And lastly you are adding one script that runs the pipeline, assuming the pipeline is there and adding the output. I wonder to which extent this belongs here. We have a tag grants script here to not rely on running the pipeline. If you think running the pipeline is the "de facto" way, this probably belongs to the pipeline repo. What is the use case?

nsorros:
⚠️ Note that I switched to main for the base branch. I switched the base branch on the PR but for tests to work you need to cherry pick 469b116038bedae879f1ab7068832d198cf12f1d. Just run `git cherry-pick 469b116038bedae879f1ab7068832d198cf12f1d` if you aim to do more commits and want to run tests against latest main and not the soonoutdated master

data/interim/mesh_pipeline_result.csv.dvc
aCampello: I think this is a mistake, I will have a look

aCampello:
> And lastly you are adding one script that runs the pipeline, assuming the pipeline is there and adding the output. I wonder to which extent this belongs here. We have a tag grants script here to not rely on running the pipeline. If you think running the pipeline is the "de facto" way, this probably belongs to the pipeline repo. What is the use case?

It's a tricky one. I want to do evaluation with _exactly_ the same dataset that is produced in prod, and I don't want to copy-paste the code in this repository here. So the dvc dag needs one dataset which is generated by an external pipeline, (`trigger_pipeline.sh`) which I need to add manually, and then it needs to do some ML logic.

Ideally, the `prod` code wouldn't contain any data processing logic, it would just call the code in here, but this is another discussion I think.

data/interim/mesh_pipeline_result.csv.dvc
aCampello: No that is correct. This is generated by `trigger_pipeline.sh` which depends on an external repository, so at the moment I am generating it manually.

nsorros:
> > And lastly you are adding one script that runs the pipeline, assuming the pipeline is there and adding the output. I wonder to which extent this belongs here. We have a tag grants script here to not rely on running the pipeline. If you think running the pipeline is the "de facto" way, this probably belongs to the pipeline repo. What is the use case?
> 
> It's a tricky one. I want to do evaluation with _exactly_ the same dataset that is produced in prod, and I don't want to copy-paste the code in this repository here. So the dvc dag needs one dataset which is generated by an external pipeline, (`trigger_pipeline.sh`) which I need to add manually, and then it needs to do some ML logic.
> 
> Ideally, the `prod` code wouldn't contain any data processing logic, it would just call the code in here, but historically this is how it is, so I think it's another discussion.

Well i did make that suggestion to Jeff in #73. Its worth reading what he said.

In any case if you want to run that pipeline just also add instructions on how a user can run it somewhere i.e. the steps to get the code in, something like git clone into that path etc etc

aCampello:
> In any case if you want to run that pipeline just also add instructions on how a user can run it somewhere i.e. the steps to get the code in, something like git clone into that path etc etc

I think that's a good idea.

data/interim/mesh_pipeline_result.csv.dvc
nsorros: And are those grants? Maybe the name could be more descriptive but I understand this might require you to rerun. Anyway if you do maybe `mesh_pipeline_tagged_grants.csv`

scripts/README.md
```diff
@@ -0,0 +1,25 @@
+## ✍️ Scripts
+
+This folder contains convenience functions to run certain tasks. It also contains
+scripts to runs pipelines which are outside this repository, some of them in 
+private Wellcome repositories.
+
+### Evaluation dag & getting grants
+The main dvc DAG has an evaluation independent dag for one of the models (MeSH). To run this,
+you need to be on the Wellcome VPN, you need to install private requirements
```
nsorros, line 9: do you mean part of WellcomeML org in Github?

scripts/generate_validation_data_xlinear.py
```diff
@@ -0,0 +1,72 @@
...
+    from datascience.warehouse.warehouse import FortyTwo
+    from datascience.grants.cleaning import clean_grants
+except ImportError as e:
+    raise ImportError(f"Error importing {e.name }. "
+                      f"To use this script you need to install Wellcome's datascience internal "
+                      f"utils, for example with `make install-private-requirements`")
+
+
+# Name of the file output of the pipeline
+predictions_file = 'mesh_pipeline_result.csv'
```
nsorros, line 16: it would be better if these were arguments to a CLI

scripts/generate_validation_data_xlinear.py
```diff
@@ -0,0 +1,72 @@
...
+
+# Name of the file with the predicitons, alongside metadata
+merged_predictions_file = 'merged_mesh_predictions_mesh_xlinear_for_validation.xlsx'
```
nsorros, line 19: same

scripts/generate_validation_data_xlinear.py
```diff
@@ -0,0 +1,72 @@
...
+grants = clean_grants(grants).fillna('')
+
+## Query FortyTwo to obtain the MeSH descriptors (could query the raw XML too?)
+descriptor_query = """
+SELECT [MeSH Lowest Level Classification], [MeSH Classification Description]
+FROM FortyTwo.dbo.MeSHClassificationHierarchy 
+"""
+
+forty_two = FortyTwo()
+descriptors = {
```
nsorros, line 32: so this script queries directly FortyTwo instead of using the results from the mesh pipeline, why?

scripts/generate_validation_data_xlinear.py
```diff
@@ -0,0 +1,72 @@
...
+predictions.rename({'Grant id': 'Grant ID'}, axis=1, inplace=True)
+
+merged_predictions_metadata = pd.merge(left=grants, right=predictions, how='right', on='Grant ID')
+merged_predictions_metadata['MeSH Classification Description'] = \
+    merged_predictions_metadata['Tag'].apply(
+        lambda x: descriptors.get(x, "")
+)
+
+# Only End Date > 2016 (or active)
+merged_predictions_metadata = \
```
nsorros, line 51: this could be an argument

scripts/generate_validation_data_xlinear.py
```diff
@@ -0,0 +1,72 @@
...
+    merged_predictions_metadata[merged_predictions_metadata['End Date'] >= '2016']
+
+# Undersample rejected applications for validation
+merged_predictions_accepted = \
+    merged_predictions_metadata[merged_predictions_metadata['Is Awarded?'] == 'Y']
```
nsorros, line 56: same

nsorros (approved):
Don't want to block this. But there are three areas you might want to think

- Add a CLI to scripts instead of hardcoding. The fact thats its a script does not mean that you cannot pass in arguments.
- Whether it should be part of the main DAG
- The name of the file that comes from running the pipeline and why we dont use that and instead fetch from 42

aCampello:
> * Whether it should be part of the main DAG

Yep, maybe. The question is what I'm using the dag for. I don't think a general user of the package has to fiddle with the dag whatsoever, so no risk in having stuff there that only we can run. Very similar to the other evaluation tasks btw, that require access to private Wellcome S3.

> * The name of the file that comes from running the pipeline and why we dont use that and instead fetch from 42

That's because the result of the pipeline only returns Grant_ID + Classification, but we need to join this with some extra columns with extra metadata to generate the file for evaluation.

scripts/generate_validation_data_xlinear.py
aCampello: I take the point of this being arguments, but I don't think I'll run this in any other file, other than the _exact_ file in `data/raw` or `data/interim` so I don't really want this name to change.

dvc.yaml
```diff
@@ -150,3 +150,18 @@ stages:
     metrics:
     - results/mesh_xlinear_on_grants.json:
         cache: false
+
+  get_grants:
```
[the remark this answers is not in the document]
aCampello: Yep, I understand the worry, but it's an independent component of the dag, so `dvc repro` will only be outdated if we change those specific scripts.

nsorros:
> > * Whether it should be part of the main DAG
> 
> Yep, maybe. The question is what I'm using the dag for. I don't think a general user of the package has to fiddle with the dag whatsoever, so no risk in having stuff there that only we can run. Very similar to the other evaluation dags btw, that require access to private Wellcome S3.
> 
> > * The name of the file that comes from running the pipeline and why we dont use that and instead fetch from 42
> 
> That's because the result of the pipeline only returns Grant_ID + Classification, but we need to join this with some extra columns with extra metadata to generate the file for evaluation.

👍

scripts/README.md
aCampello: Nope, I mean Wellcome VPN, to query 42. You will also need to be part of the org

PR #173, opened 2021-12-07
Add Xtransformer experiments

nsorros:
## Description

Related to #120 #121 #124 #125 

## Checklist

- [x] Linked to Notion or GitHub issue
- [ ] Added tests
- [ ] Updated README
- [ ] DVC up to date

## Release checklist

- [ ] DVC repro up to date
- [ ] Models synced in S3
- [ ] Version updated
- [ ] CHANGELOG.md updated
- [ ] Model and package version aligned

To release:

* `make build`
* `make deploy`

PR #160, opened 2021-11-24
Add Bert mesh experiments 

nsorros:
## Description

Fixes #127 

## Checklist

- [x] Linked to Notion or GitHub issue
- [ ] Added tests
- [ ] Updated README
- [ ] DVC up to date

PR #174, opened 2021-12-07
Add year, journal in preprocess mesh

### LINKED ISSUE #109: Add year in meta in grants_tagger and add filter by year param to preprocess_mesh
https://github.com/wellcometrust/grants_tagger/blob/720bdc46ff3f7b09407d8d01825450a67742bad6/grants_tagger/preprocess_mesh.py#L86

`preprocess_mesh` should add a parameter `--filter-years` that accept a comma separated `min_year,max_year` to filter the raw data.

this way the user can create train and test sets with a particular set of publication years which seems to have a big effect in training

nsorros:
## Description

Both #173 and #160 require a subset of the data to train. The best possible way to provide that subset is to filter by years which is standard practice in this problem. At the moment this is implemented in both PR mentioned but this aims to resolve it for both.

Fixes #109

## Checklist

- [x] Added tests
- [x] Updated README
- [ ] DVC up to date

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,20 +89,31 @@ Options:
 #### bioasq-mesh
 ```
 Usage: grants_tagger preprocess bioasq-mesh [OPTIONS] [INPUT_PATH]
-                                            [OUTPUT_PATH]
+                                            [TRAIN_OUTPUT_PATH]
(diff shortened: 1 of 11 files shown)


pdan93 (approved):
LGTM :+1:

Issue #154, opened 2021-11-15
Fix train fail in tox


nsorros:
The train command fails with tox but not with pytest which is weird. The error is
```
def test_train_command():
        with tempfile.TemporaryDirectory() as tmp_dir:
            data_path = os.path.join(tmp_dir, "data.jsonl")
            model_path = os.path.join(tmp_dir, "model.pkl")
            label_binarizer_path = os.path.join(tmp_dir, "label_binarizer.pkl")
            train_info_path = os.path.join(tmp_dir, "train_info.json")

            write_jsonl(data_path, DATA)

            result = runner.invoke(app, [
                "train",
                data_path,
                model_path,
                label_binarizer_path,
                "--approach",
                "tfidf-svm",
                "--parameters",
                "{'tfidf__min_df': 1, 'tfidf__stop_words': None}",
            #    "--train-info",
            #    train_info_path
            ])
>           assert result.exit_code == 0
E           AssertionError: assert 1 == 0
E            +  where 1 = <Result FileNotFoundError(2, 'No such file or directory')>.exit_code
```

For now we mark the test and skip for tox but it would be nice to resolve this

nsorros:
This is fixed as part of #175

PR #175, opened 2021-12-09
Refactor cli

### LINKED ISSUE #117: Create CLI in cli.py and in each script
At the moment we have the CLI defined in `__main__.py` to allow for the following use `python -m grants_tagger.train` which I personally never use but regardless we can define it more cleanly in its own file `cli.py` and then import the `app` in `__main__.py`.

Additionally each script should have its own cli that someone can use for example `python grants_tagger/train.py` and that CLI should be imported inside cli and assembled into the main cli we expose.

pdan93:
## Description
Move CLI logic from __main__.py to cli.py. Have every script declare it's own cli, and the main cli.py just imports those for the complete CLI

Fixes #117 

## Checklist

- [x] Added tests
- [ ] Updated README
- [ ] DVC up to date


grants_tagger/cli.py
```diff
@@ -0,0 +1,142 @@
...
+from grants_tagger.tune_threshold import tune_threshold_cli
+from grants_tagger.optimise_params import tune_params_cli
+from grants_tagger.download_epmc import download_epmc_cli
+from grants_tagger.download_model import download_model_cli
+from grants_tagger.explain import explain_cli
+
+app = typer.Typer(add_completion=False)
+
+
+# Move to train and import from there
```
nsorros, line 39: we can remove that now

nsorros:
⚠️ Don't forget to include the app in `__main__` so `python -m grants_tagger` works

tests/test_cli.py
```diff
@@ -133,19 +134,16 @@ def test_train_command():
         result = runner.invoke(app, [
             "train",
             data_path,
-            model_path,
             label_binarizer_path,
+            model_path,
```
nsorros, line 138: did we change the order? this might affect `dvc.yaml`

pdan93: resolving :)

grants_tagger/train_with_sagemaker.py
```diff
@@ -141,3 +148,100 @@ def train_with_sagemaker(instance_type="local", config_version=None, **kwargs):
...
+        threshold=threshold,
+        data_format=data_format,
+        sparse_labels=sparse_labels,
+        cache_path=cache_path,
+        instance_type=instance_type,
+        config_version=config_version,
+    )
+
+    duration = time.time() - start
+    print(f"Took {duration:.2f} to train")
```
nsorros, line 238: This was supposed to measure how much time train took. It should go back to the main train in cli.py

nsorros (changes requested):
There are two things that need attention

- ensuring `dvc.yaml` works
- the training time info should go back to main train

Other than that, its looking good and its ready to be merged

tests/test_cli.py
pdan93: you know what the weird thing is, dvc actually had the correct order, just the test had the wrong one. Meaning it was passing until now with the wrong order somehow

pdan93: @nsorros

grants_tagger/train_with_sagemaker.py
pdan93: Reason I moved it in both train scripts is so it measures it also when you call the individual scripts, doesn't it make more sense like that?

pdan93:
> warning Don't forget to include the app in `__main__` so `python -m grants_tagger` works

already included

pdan93:
With last commit I verified order of all arguments in dvc.yaml and there was only one change in the `preprocess_wellcome_science` step. Everything else seems in order, but I have no way of running `dvc repro`, you'll have to do it @nsorros

grants_tagger/train_with_sagemaker.py
nsorros: Actually it mostly makes sense for train that happens locally so remove it from sagemaker and only have it in train_cli. Also in train_cli you do not pass the train_info. We want to save this information in a file and you are only printing in screen

tests/test_cli.py
```diff
@@ -133,19 +134,16 @@ def test_train_command():
...
             "--approach",
             "tfidf-svm",
             "--parameters",
             "{'tfidf__min_df': 1, 'tfidf__stop_words': None}",
-            "--train-info",
```
nsorros, line 142: part of the reason you are not catching the problem i mentioned above is you remove that test. btw there is a problem with that test see #154. could we also fix that as part of this PR

tests/test_cli.py
pdan93: I don't get it, it shouldn't have based on what's in the train code.

nsorros (approved):
This looks ready but let me run dvc repro overnight to iron any issues that we did not find in the code review. Also remove the time info from sagemaker.

pdan93:
> om sagemaker. And resolve the conflict

done and done :)

PR #177, opened 2021-12-14
Save sparse weights in MeshTfidfSVM

### LINKED ISSUE #114: Save sparse weights in mesh tfidf-svm
`SGDClassifier` https://scikit-learn.org/stable/modules/generated/sklearn.linear_model.SGDClassifier.html which is what we use under the hood for `MeshTfidfSVM` provides a sparsity function that creates sparse weights which results in a much smaller model size. To effectively use that, it would be better to zero weights below a value say `0.001` that do not contribute much.

As a reminder the disease mesh model ended up being 70GB and if we trained on the entire mesh it would hundreds of GB. Weight pruning is a commonly used strategy to reduce model size.

Note that to use the weights for predictions you need to make them `dense` through `densify` so some care is still needed to fit into memory. Batching predictions by label as we do now will still be a good way to deal with that.

nsorros:
## Description

This helps us reduce size in the same that XLinear does and that I described in the blog post. Eventually MeshTfidfSVM might become an in house alternative to XLinear.

Fixes #114

## Checklist

- [ ] Added tests
- [ ] Updated README
- [ ] DVC up to date

### DIFF
--- a/grants_tagger/models/mesh_tfidf_svm.py
+++ b/grants_tagger/models/mesh_tfidf_svm.py
@@ -14,13 +14,19 @@ from grants_tagger.models.utils import get_params_for_component
 
 class MeshTfidfSVM:
     def __init__(
-        self, y_batch_size=256, nb_labels=None, model_path=None, threshold=0.5
+        self,
+        y_batch_size=256,
+        nb_labels=None,
+        model_path=None,
+        threshold=0.5,
+        cutoff_prob=0.1,
     ):
         """
         y_batch_size: int, default 256. Size of column batches for Y i.e. tags that each classifier will train on
         nb_labels: int, default None. Number of tags that will be trained.
         model_path: path, default None. Model path being used to save intermediate classifiers
         threshold: float, default 0.5. Threshold probability on top of which a tag is assigned
+        cutoff_prob: float, default 0.1. Threshold below which to zero the weights to reduce size by sparsifying
 
         Note that model_path needs to be provided as it is used to save
         intermediate classifiers trained to reduce memory usage.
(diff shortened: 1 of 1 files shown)


grants_tagger/models/mesh_tfidf_svm.py
```diff
@@ -82,7 +89,11 @@ def fit(self, X, Y):
...
             self.classifier.fit(X_vec, Y[:, tag_i : tag_i + self.y_batch_size])
 
-            # TODO: Sparsify weights before saving
+            for i, est in enumerate(self.classifier.estimators_):
+                est.coef_[np.abs(est.coef_) < self.cutoff_prob] = 0
+                est.sparsify()
+                self.classifier.estimators_[i] = est
+
             with open(f"{self.model_path}/{tag_i}.pkl", "wb") as f:
                 f.write(pickle.dumps(self.classifier))
```
ivyleavedtoadflax, line 98: Is there a reason for using `pickle.dumps()` rather than `pickle.dump`. If you use the latter, you don't need `f.write()`:

```
            with open(f"{self.model_path}/{tag_i}.pkl", "wb") as f:
                pickle.dump(self.classifier, f)
```

> pickle.dump(obj, file, protocol=None, *, fix_imports=True, buffer_callback=None)¶
> Write the pickled representation of the object obj to the open file object file. This is equivalent to Pickler(file, ?protocol).dump(obj).
>
> Arguments file, protocol, fix_imports and buffer_callback have the same meaning as in the Pickler constructor.
>
> Changed in version 3.8: The buffer_callback argument was added.

https://docs.python.org/3/library/pickle.html

ivyleavedtoadflax (approved):
Looks good. Just one comment. :shipit:

grants_tagger/models/mesh_tfidf_svm.py
nsorros: I am defaulting to this way for a while as its easier to remember for me. In other cases I was always confusing whether it was dump or dumps. Now I only use `loads` and `dumps` explicitly

PR #179, opened 2021-12-14
Return sparse probs in mesh models

### LINKED ISSUE #113: Return sparse probs in mesh models
`MeshXLinear` returns sparse probabilities already by not including probs below a certain value, say 0.01. We can do the same for `MeshCNN` and `MeshTfidfSVM`. In fact I did try including this the PR #108 but some tests were failing as the code currently expects a dense numpy arrays so we need to update certain pieces mainly in `tune_threshold` and `explain`. For explain, it might be slightly more complicated because we depend on an external library `shap` which needs updating but one solution is to drop supporting explanations until this is resolved and open an issue to resolve it later or we can pass a flag that returns sparse_probs and set it to False for `explain`.

nsorros:
## Description

Fixes #113

## Checklist

- [ ] Added tests
- [ ] Updated README
- [ ] DVC up to date

### DIFF
--- a/grants_tagger/models/mesh_cnn.py
+++ b/grants_tagger/models/mesh_cnn.py
@@ -20,6 +20,7 @@ class MeshCNN:
         buffer_size=1000,
         data_cache=None,
         random_seed=42,
+        cutoff_prob=0.1,
     ):
         """
         threshold: float, default 0.5. Probability threshold on top of which a tag should be assigned.
@@ -28,6 +29,7 @@ class MeshCNN:
         buffer_size: int, default 1000. Buffer size used for shuffling or transforming data before fit.
         data_cache: path, default None. Path to use for caching data transformations.
         random_seed: int, default 42. Random seed that controls reproducibility.
+        cutoff_prob: float, default 0.1. Prob below which we zero the probs to create sparse probs
         """
         self.threshold = threshold
         self.batch_size = batch_size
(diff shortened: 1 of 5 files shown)


grants_tagger/models/mesh_cnn.py
```diff
@@ -20,6 +20,7 @@ def __init__(
         buffer_size=1000,
         data_cache=None,
         random_seed=42,
+        cutoff_prob=0.1,
```
ivyleavedtoadflax, line 23: A philosophical question. Should you set defaults here at all?

tests/test_predict.py
```diff
@@ -263,6 +263,7 @@ def test_predict_tags_mesh_cnn(mesh_cnn_path, mesh_label_binarizer_path):
     tags = predict_tags(
         X, mesh_cnn_path, mesh_label_binarizer_path,
         approach="mesh-cnn")
+    print(tags)
```
ivyleavedtoadflax, line 266: Rogue print?

ivyleavedtoadflax (approved):
Looks good. I didn't run it, but it all makes sense.

grants_tagger/models/mesh_cnn.py
nsorros: Given that otherwise Y_pred_proba does not fit to memory most of the time, this way it does and additionally most of low probs do not add any value.

Additionally sklearn has this philosophy of always having default if this is purely philosophical so you can initialize all classes by simply calling them e.g. SGDClassifier() or SVC()

tests/test_predict.py
nsorros: 😬 this should go

nsorros: 2e0218b

PR #182, opened 2022-01-19
Add download, tag and submit scripts for bioasq

nsorros:
## Description

Add scripts to download, tag and submit data to the BioASQ competition which automates an otherwise manual process.

## Checklist

- [ ] Added tests
- [ ] Updated README
- [ ] DVC up to date

## Release checklist

- [ ] DVC repro up to date
- [ ] Models synced in S3
- [ ] Version updated
- [ ] CHANGELOG.md updated

To release:

* `make build`
* `make deploy`

### DIFF
--- a/scripts/tag_bioasq_test_data.py
+++ b/scripts/tag_bioasq_test_data.py
@@ -1,0 +1,18 @@
+import typer
+
+from grants_tagger.predict import predict_tags
+
+
+def tag_bioasq_data(data_path, model_path, label_binarizer_path, tagged_data_path, threshold=0.5):
+    data = json.load(data_path)
+    X = [item["Title"] + " " + item["abstractText"] for item in data]
+    pmids = [item["pmid"] for item in data]
+
+    tags = predict_tags(X, model_path, label_binarizer_path, "mesh-xlinear", threshold=threshold)
+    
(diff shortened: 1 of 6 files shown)


scripts/tag_bioasq_test_data.py
```diff
@@ -0,0 +1,45 @@
...
+    mesh2code = {}
+    for mesh in tqdm(mesh_tree.getroot()):
+        try:
+            # DescriptorUI e.g. M000616943
+            mesh_code = mesh[0].text
+            # DescriptorName e.g. Mucosal-Associated Invariant T Cells
+            mesh_name = mesh[1][0].text
+            
+            mesh2code[mesh_name] = mesh_code
+        except IndexError:
```
aCampello, line 21: Why would IndexError happen, considering the name is correct ? Would be good to investigate further how many errors happened (maybe print in a log?)

aCampello (approved):
Very neat and clean, thanks a lot for this PR Nick.

scripts/tag_bioasq_test_data.py
nsorros: This is copy from https://github.com/wellcometrust/grants_tagger/blob/main/grants_tagger/filter_mesh_tags.py

I can definitely add logger or we can also refactor the code in the other function. That said, scripts in theory should be independent, right? I think the way the code is structured its not meant to share code.

PR #189, opened 2022-02-22
Add BertMesh to models so that we can tag in BioASQ

nsorros:
## Description

Adds BertMesh in models by copying and not refactoring the experimental code in BertMesh. The reason being that at this point we want to maintain moving fast in experimentation while separating what we maintain for "production". As such we do not add code to train BertMesh as this happens still in experimentation mode. Once this becomes more stable we can do that.

DVC sort of up to date.

## Checklist

- [x] Added tests
- [ ] Updated README
- [x] DVC up to date

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,7 +12,7 @@ SCIENCE_TFIDF_SVM_MODEL:= tfidf-svm.pkl
 SCIENCE_SCIBERT_MODEL := scibert
 SCIENCE_LABEL_BINARIZER := label_binarizer.pkl
 
-PYTHON := python3.8
+PYTHON := python3.7
 VIRTUALENV := venv
 PIP := $(VIRTUALENV)/bin/pip
 
--- a/Makefile
+++ b/Makefile
@@ -12,7 +12,7 @@ SCIENCE_TFIDF_SVM_MODEL:= tfidf-svm.pkl
 SCIENCE_SCIBERT_MODEL := scibert
 SCIENCE_LABEL_BINARIZER := label_binarizer.pkl
 
-PYTHON := python3.7
+PYTHON := python3
 VIRTUALENV := venv
 PIP := $(VIRTUALENV)/bin/pip
 
--- a/grants_tagger/models/bert_mesh.py
+++ b/grants_tagger/models/bert_mesh.py
@@ -21,10 +21,28 @@ class MeshDataset(Dataset):
 
 
 class WellcomeBertMesh:
-    def __init__(self, cutoff_prob=0.1, threshold=0.5, batch_size=16):
+    def __init__(
+        self,
+        cutoff_prob=0.1,
+        threshold=0.5,
+        batch_size=16,
+        pretrained_model="microsoft/BiomedNLP-PubMedBERT-base-uncased-abstract",
+        num_labels=28761,
+        hidden_size=1024,
+        multilabel_attention=True,
+    ):
         self.cutoff_prob = cutoff_prob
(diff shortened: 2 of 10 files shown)


Makefile
```diff
@@ -12,7 +12,7 @@ SCIENCE_TFIDF_SVM_MODEL:= tfidf-svm.pkl
 SCIENCE_SCIBERT_MODEL := scibert
 SCIENCE_LABEL_BINARIZER := label_binarizer.pkl
 
-PYTHON := python3.8
+PYTHON := python3.7
```
nsorros, line 15: this goes back and forth according to the environment, maybe we should revert to python3 to not have to change often and ensure the code runs for both versions

aCampello: Agree. or an env variable that the user can define? I keep changing to 3.8 as this is the one installed in my machine currently

grants_tagger/models/bert_mesh.py
```diff
@@ -0,0 +1,77 @@
...
+import scipy.sparse as sp
+import numpy as np
+import torch
+
+from grants_tagger.bertmesh.model import BertMesh
+
+
+class MeshDataset(Dataset):
+    def __init__(self, X):
+        self.tokenizer = BertTokenizerFast.from_pretrained("bert-base-uncased")
```
aCampello, line 13: Should the tokeniser size be a parameter?

I think it's called `max_position_embeddings`

grants_tagger/models/bert_mesh.py
```diff
@@ -0,0 +1,77 @@
...
+
+        self.model.eval()
+
+        Y_pred_proba = []
+        with torch.no_grad():
+            for inputs in tqdm(data):
+                outs = self.model(inputs)
+
+                Y_pred_proba_batch = outs.cpu().numpy()
+                Y_pred_proba_batch[Y_pred_proba_batch < self.cutoff_prob] = 0
```
aCampello, line 62: I don't understand why you do this.

grants_tagger/models/bert_mesh.py
```diff
@@ -0,0 +1,77 @@
...
+                Y_pred_proba.append(sp.csr_matrix(Y_pred_proba_batch))
```
aCampello, line 63: Ah, it's to use the sparse matrix trick. Nice

aCampello (approved):
The class is really nice, thanks for doing this. Can I be a massive pain and ask that you either add somewhere in the documentation the actual script/steps to tag (like the parameters passed to `python scripts/tag_bioasq_test_data` ? ) or add a dvc pipeline for that. Your choice.

nsorros:
> The class is really nice, thanks for doing this. Can I be a massive pain and ask that you either add somewhere in the documentation the actual script/steps to tag (like the parameters passed to `python scripts/tag_bioasq_test_data` ? ) or add a dvc pipeline for that. Your choice.

Sure this is a good idea. Its kind of hacky at the moment.

grants_tagger/models/bert_mesh.py
nsorros: Yep, we don't need all the probs, let's reduce the size.

grants_tagger/models/bert_mesh.py
nsorros: what do you mean? I am not familiar with it

nsorros:
This is ready to go. I will run the dvc pipeline on Monday to ensure it works smoothly in the latest batch and merge.

PR #191, opened 2022-02-25
BertMesh meets transformers

nsorros:
## Description

Even though we have upload BertMesh model into the hugging face hub [here](https://huggingface.co/Wellcome/WellcomeBertMesh) the weights are not initialised properly and additionally making it work required a couple of manual tweaks.

A more standard way to get a custom model into the hub is to subclass `PreTrainedModel` instead of `torch.nn.Module` which gives you access to `save_pretrained` and `load_pretrained`. It seems that `PreTrainedModel` is itself a wrapper on top `nn.Module` with some extras so it should not restrict us too much hopefully 🤞 

This is ready to run @aCampello. Note you need to recreate the environment
```
make virtualenv
dvc repro grants_tagger/bertmesh/dvc.yaml
```
Also you need to get weights and biases creds from Nick. This needs to run in `g4dn.metal` 🔥 

## Checklist

- [x] Added tests
- [ ] Updated README
- [ ] DVC up to date


aCampello:
Happy to run this, just let me know how to set up the credentials

nsorros:
One thing that needs to happen is to update to the latest WellcomeML after https://github.com/wellcometrust/WellcomeML/pull/384 is merged which will allow us to get the latest version of transformers which enables one functionality needed to make it easier to upload to the hub.

aCampello:
It seems to me that the model on the model hub is still the old one:

https://huggingface.co/Wellcome/WellcomeBertMesh/tree/main

aCampello:
The hub is also a bit weird, it says to use it we need to:

```bash
from transformers import AutoTokenizer, BertMesh

tokenizer = AutoTokenizer.from_pretrained("Wellcome/WellcomeBertMesh")

model = BertMesh.from_pretrained("Wellcome/WellcomeBertMesh")
```

Should it be BertModel instead?

nsorros:
> The hub is also a bit weird, it says to use it we need to:
> 
> ```shell
> from transformers import AutoTokenizer, BertMesh
> 
> tokenizer = AutoTokenizer.from_pretrained("Wellcome/WellcomeBertMesh")
> 
> model = BertMesh.from_pretrained("Wellcome/WellcomeBertMesh")
> ```
> 
> Should it be BertModel instead?



> It seems to me that the model on the model hub is still the old one:
> 
> https://huggingface.co/Wellcome/WellcomeBertMesh/tree/main

Why do you say that?

nsorros:
Labels added in config so now you can get labels straight out of the model. Try it
```
from transformers import AutoModel, AutoTokenizer

tokenizer = AutoTokenizer.from_pretrained("Wellcome/WellcomeBertMesh")
model = AutoModel.from_pretrained("Wellcome/WellcomeBertMesh", trust_remote_code=True)

inputs = tokenizer(["This grant is about malaria and not about HIV."], padding="max_length")
labels = model(**inputs, return_labels=True)
print(labels)
```

pdan93 (commented):
I think we forgot one thing. batch size. This code I think would work only for bz=1

grants_tagger/bertmesh/model.py
```diff
@@ -55,4 +56,7 @@ def forward(self, input_ids, **kwargs):
             outs = torch.nn.functional.relu(self.linear_1(cls))
             outs = self.dropout_layer(outs)
             outs = torch.sigmoid(self.linear_out(outs))
+        if return_labels:
+            # TODO Vectorize
+           outs =  [[self.id2label[label_id] for label_id, label_prob in enumerate(out) if label_prob > 0.5] for out in outs]
```
pdan93: I think this might not work for batch size > 1 ?

aCampello:
I can confirm that this works for me now

grants_tagger/bertmesh/model.py
nsorros: Hmm, I can test this but why would it not work? `for out in outs` should handle that

nsorros:
> I can confirm that this works for me now

Did you approve as well? :D

grants_tagger/bertmesh/model.py
```diff
@@ -15,42 +15,55 @@ def forward(self, x):
...
             outs = torch.nn.functional.relu(self.linear_1(cls))
             outs = self.dropout_layer(outs)
             outs = torch.sigmoid(self.linear_out(outs))
+        if return_labels:
+            # TODO Vectorize
+            outs = [
+                [
+                    self.id2label[label_id]
+                    for label_id, label_prob in enumerate(out)
+                    if label_prob > 0.5
```
aCampello, line 65: Add threshold as a parameter perhaps?

grants_tagger/bertmesh/train_torch.py
```diff
@@ -178,12 +209,22 @@ def train_bertmesh(
...
+        model = model.module
+
+    model.save_pretrained(model_path)
 
     if train_metrics_path:
         with open(train_metrics_path, "w") as f:
             f.write(json.dumps({"metrics": metrics}))
 
+    duration = time.time() - start
+    instance = get_ec2_instance_type()
```
aCampello, line 222: 👌

aCampello (approved):
Happy to go ahead with that if you fix black :) Thanks so much, and sorry for the delay!

PR #192, opened 2022-02-25
Feature/new xlinear model

### LINKED ISSUE #193: Onlytop_k not working
When predicting `mesh-xlinear` with `only_topk` the value of 20 is always used, regardless of which value is set up. Therefore we can only output 20 predictions for each datapoint, which is problematic in high recall set-ups.

The reason for that is because this parameter should be passed to prediction.

aCampello:
## Description

Fixes #193, where only_topk was not making any effect on how many predictions.

The PR has many lines because it changes the validation/exploration notebook.

## Checklist

- [ ] Added tests
- [ ] Updated README
- [x] DVC up to date

## Release checklist

- [ ] DVC repro up to date
- [ ] Models synced in S3
- [ ] Version updated
- [ ] CHANGELOG.md updated

To release:

* `make build`
* `make deploy`


grants_tagger/models/mesh_xlinear.py
```diff
@@ -138,7 +146,11 @@ def save(self, model_path):
         self.xlinear_model_.save(model_path)
 
     def load(self, model_path, is_predict_only=True):
+        params_path = os.path.join(model_path, "params.json")
```
nsorros, line 149: so params were not saved and loaded using their save / load function? Did you raise an issue with them as to why? this might be a bug that need to be aware of or that you can fix on their side

results/mesh_xlinear.json
```diff
@@ -1 +1 @@
-[{"threshold": "0.50", "precision": "0.74", "recall": "0.41", "f1": "0.53"}]
\ No newline at end of file
+[{"threshold": "0.20", "precision": "0.62", "recall": "0.52", "f1": "0.56"}]
```
nsorros, line 1: 💥

nsorros (approved):

grants_tagger/models/mesh_xlinear.py
aCampello: I think their parameters are saved fine, it's just the parameters of our class, and since we need some of the parameters for predictions, we needed to save it too.

PR #213, opened 2022-04-11
Add precommit hooks

nsorros:
## Description

Adds [precommit](https://pre-commit.com) hooks to the project so that we can share hooks and also ensure
- code is formatted with black
- dvc hooks are installed
- checks for yaml files and common linting problems happen

The changes here affect DVC so a `dvc commit` probably needed to resolve status

## Checklist

- [ ] Added tests
- [ ] Updated README
- [ ] DVC up to date


ivyleavedtoadflax:
How come this failed in a black test. I thought the pre-commit hook would fix this? :thinking:

ivyleavedtoadflax (approved):
In any case, happy to approve pending the test fix :rocket: 

Could it be a black version issue?
.github/workflow: black==22.3.0 
requirements.txt: black==21.7b0

In any case these two should probably be aligned?

nsorros:
> In any case, happy to approve pending the test fix 🚀
> 
> Could it be a black version issue? .github/workflow: black==22.3.0 requirements.txt: black==21.7b0
> 
> In any case these two should probably be aligned?

It has to do with these versions indeed. There is a click error for version `21.7b0` which I have forced the action to use click `8.0.0` but not sure how to do that in `pre-commit` that uses a separate environment to run. So I thought I would use the latest black which does not have that error but then there is another problem which is that the action and precommit disagree one some rules.

One way to fix that is to remove black from the project `unpinned` and only run it through `pre-commit` and action which means we can use the latest black in both.

ivyleavedtoadflax:
Good idea. LGTM :shipit:

nsorros:
@aCampello any thoughts on this?

PR #230, opened 2022-11-17
dvc hash of mesh terms used by WT

ArneRobben:
## Description
Just a dvc hash file which has the location of a text file with all mesh terms that are used by WT


ArneRobben:
Thanks Antonia!

PR #233, opened 2022-11-25
PR to retrain MeSH classifier

ArneRobben:
## Description
This PR is for the retraining of the MeSH classifier. It's a pretty bulky repository (given its history), but hopefully the changes aren't too bad..

The main additions are: 
- a script `create_inclusion_list.py` which effectively excludes terms we do not want to use in our tagging anymore:
  - tags that are being tagged in the MeSH dictionary with a 'DO NOT USE' attribute. These are MeSH tags that are there because they help catalogue or help with structuring the tree or are contentious such as Male and Female. Theres 200 of these MeSH tags
  - I also included a file `descriptors_not_to_use_manual.csv` in data/processed where we can add labels that will be excluded from training. E.g. Kirstin asked me to take out Humans, which is the only term in there so far.
 - an updated `dvc.yaml` file which contains three pipelines in one file: one for retraining the MeSH classifier using the above inclusion list, one for retraining the MeSH classifier on terms used by WT (which has slightly better accuracy) and one with toy data, because running any of the two other pipelines takes a while.

## Checklist

- [x] [Linked to Notion](https://www.notion.so/wellcometrust/Retrain-MeSH-classifier-reduced-list-b80d0fbbeec14912a8f45d78cc2c13ed)
- [ ] Added tests
- [x] Updated README
- [x] DVC up to date

## Release checklist

- [x] DVC repro up to date
- [x] Models synced in S3
- [x] Version updated (to do)
- [x] CHANGELOG.md updated (to do)
- [x] Model and package version aligned (to do)

To release:

* `make build`
* `make deploy`


README.md
```diff
@@ -79,6 +79,16 @@ preprocessing so the current preprocess works with the wellcome-science dataset
 the bioasq-mesh one. If you want to use a different dataset see section on bringing
 your own data under development.
 
+### inclusion list
```
Amvla, line 82: Nice explanation, great to have this in the README 👍

grants_tagger/create_inclusion_list.py
```diff
@@ -0,0 +1,66 @@
...
+import os
+import httpx
+import uuid
+import xml.etree.ElementTree as ET  # this is a library that can parse xml easily
+import pandas as pd
+from pathlib import Path
+
+
+def inclusion(xml_path, out_path):
+
```
Amvla, line 17: Could you also add a docstring to the function? I can see you have a nice explanatory comment at the top, but it would be good to specifically explain your function as well.

grants_tagger/create_inclusion_list.py
```diff
@@ -0,0 +1,66 @@
...
+            annotations.append("")
+        else:
+            annotations.append(mesh.find("Annotation").text)
+
+    descriptors_to_use = []
+    descriptors_set = zip(descriptors, annotations)
+
+    # read in the list of descriptors we would like to drop manually:
+    exclusion_df = pd.read_csv(exclusion_list_path)
+    exclusion_list = list(exclusion_df["DescriptorName"].values)
```
Amvla, line 37: Could potentially also use pandas method `exclusion_df["DescriptorName"].tolist()`

grants_tagger/create_inclusion_list.py
```diff
@@ -0,0 +1,66 @@
...
+
+    # make a list of the descriptors we want to use
+    for descriptor, annotation in descriptors_set:
+        if (
+            "Do not use" not in annotation and descriptor not in exclusion_list
```
Amvla, line 42: Does "Do not use" always start with a capital D? If you wanted to make sure and catch all cases, you could check "do not use" against annotation.lower()

results/results.json
```diff
@@ -0,0 +1 @@
+{"threshold": "0.50", "precision": "0.74", "recall": "0.41", "f1": "0.53"}
```
Amvla, line 1: Just a quick question, which part of the dvc.yaml are these results (and the ones below) linked to? Sorry you might have mentioned it earlier but now I can't figure out anymore what evaluation these are part of (vs. the results above)... they look good though, especially the wt ones 🙂

Amvla (changes requested):
Really nice to be able to filter the MeSH terms by both annotations and a manual list! My only small request is adding a docstring to your inclusion function, then I'm happy to approve 🙂
Please feel free to ignore my other comments/question if you feel like they're not relevant.

grants_tagger/create_inclusion_list.py
ArneRobben: Great suggestion!

grants_tagger/create_inclusion_list.py
ArneRobben: It does. In fact, I am going to make this a little stronger and make it ": Do not use" as the examples the format is usually `INDEXER: Do not use` or `CATALOGER: Do not use`

results/results.json
ArneRobben: so there's three pipelines in the yaml:
- the 'normal' xlinear one, which is the full MeSH tree, except for the c. 200 terms we are now excluding with `create_inclusion_list.py`. The evaluation results from this model are here in `results.json`
- the WT pipeline, which only uses the MeSH tags we use at Wellcome (c. 85% of the full MeSH tree), the results of which are in `results_wt.json`
- the toy pipline, which only uses a handful of MeSH tags, the results of which are in `results_toy.json`

grants_tagger/create_inclusion_list.py
Amvla: Great, thanks 🙂 love a nice docstring
I like adding the type in brackets to the arguments to make it even less ambiguous when you have lots of arguments, but pretty obvious here anyway... (I think Justin added the exact formatting to the data science portal)

Amvla (approved):

PR #238, opened 2023-01-03
hopefully fixed dependencies

ArneRobben:
## Checklist

- [ ] Added tests
- [x] Updated README
- [x] DVC up to date

## Release checklist

- [x] DVC repro up to date
- [ ] Models synced in S3
- [ ] Version updated
- [ ] CHANGELOG.md updated

To release:

* `make build`
* `make deploy`


ArneRobben:
If this fix works, it makes PR #235 redundant.

dvc.yaml
```diff
@@ -72,17 +72,17 @@ stages:
...
-      --approach mesh-xlinear --sparse-labels --train-info results/mesh_xlinear_train_info_wt.json --slim
-    deps:
-    - data/processed/train_mesh2021_wt.jsonl
-    - grants_tagger/train.py
-    - grants_tagger/slim/mesh_xlinear.py
-    params:
-    - train.mesh-xlinear.config
-    outs:
-    - models/xlinear-wt/model-${version_number}
+  # train_mesh_xlinear_wt_only:
```
Amvla, line 75: Just to check the reasoning behind commenting these bits of code... would it normally be necessary to run these parts of the pipeline but you only switched them off temporarily to check if everything is working? If it is only temporarily disabled, it might be worth uncommenting again before you merge into main

Amvla (commented):
Just a quick question around the commented bits of code, otherwise looks fine to me 🙂

dvc.yaml
ArneRobben: these bits of the pipeline create a model that works on MeSH tags that are used by Wellcome grants (c. 85%), rather than the full MeSH list (100%). We have decided in Portfolio Analysis that such a model is however not a priority so I have commented this out.. I might group these into a separate section and write a comment explaining as such

Amvla: Ok great, thanks for the explanation :) If it can be grouped into a section that would look really nice and readable

Amvla (approved):

PR #246, opened 2023-05-09
Add light grants tagger for inference as the default 🪶

### LINKED ISSUE #201: Decouple production code and experimental code/package
At the moment, to run any model (e.g. xlinear) in production, we need to install the whole package. This is probably not ideal, and blows up the size of the docker container. It would be good to have a minimal code just for the production package.

pdan93:
## Description

This PR makes the default grants tagger environment light and only responsible to inference. To do that

- we split the dependencies into `dev` and `default`
- we introduce another command for updating the `dev` requirements which is `make update-requirements-dev`
- we split tests into default and dev and run them with separate deps to ensure they work on their respective installations
- remove xlinear and cnn models and making wellcome bert mesh the default.
- make use of the cpu installation of torch that reduces the size of virtualenv from almost 6GB to 1.4GB 🔥

We also

- remove the explain command which only worked for the cnn model
- remove download model command which only made sense for xlinear

You can install the default environment and test by running
```
make virtualenv
make test
```

Fixes #201

## Checklist

- [ ] Added tests
- [ ] Updated README
- [ ] DVC up to date

## Release checklist

- [ ] DVC repro up to date
- [ ] Models synced in S3
- [ ] Version updated
- [ ] CHANGELOG.md updated


grants_tagger/module_tester.py
```diff
@@ -0,0 +1,22 @@
+import logging
+
+logger = logging.getLogger(__name__)
+
+development_dependencies = False
```
nsorros, line 5: move this file to utils. also move the import stage inside the function

grants_tagger/module_tester.py
```diff
@@ -0,0 +1,22 @@
...
+    import pecos
+    import wellcomeml
+
+    development_dependencies = True
+except ImportError:
+    logger.warning("Development dependencies not installed.")
+    pass
+
+
+def test_development_dependencies():
```
nsorros, line 16: rename to `import_development_dependencies`

Makefile
```diff
@@ -79,6 +75,31 @@ update-requirements: ## Updates requirement
...
+	virtualenv --python $(PYTHON) $(VIRTUALENV)
+	$(PIP) install --upgrade pip
+	$(PIP) install pytest pytest-cov tox
+	$(PIP) install -r dev_requirements.txt
+	$(PIP) install --no-deps -e .
+	$(PIP) install https://s3-us-west-2.amazonaws.com/ai2-s2-scispacy/releases/v0.4.0/en_core_sci_sm-0.4.0.tar.gz
+	$(VIRTUALENV)/bin/pre-commit install --hook-type pre-push --hook-type post-checkout --hook-type pre-commit
+
+update-requirements-dev: VIRTUALENV := /tmp/update-requirements-venv/
+update-requirements-dev: ## Updates requirement
```
nsorros, line 91: how can you update requirements.txt?

setup.py
```diff
@@ -10,18 +10,5 @@
     packages=find_packages(),
     version=__version__,
     entry_points={"console_scripts": "grants_tagger=grants_tagger.__main__:app"},
-    install_requires=[
```
nsorros, line 13: we still want this to act as a library but we want two set of requirements one for inference which should go here and an extra `dev` for the rest

nsorros (commented):
This is almost there. There are some minor things to fix before we merge

- We need two requirements file and a way to update them through a recipe. For example the default requirements file does not work in my mac because the dependencies need updating and there is no way to update
- We need to keep pinned dependencies in setup.py so that it can act as a library. We need default pinned dependencies and an extra `dev`. The pinned versions should match the equivalent requirement files
- Rename the test_development_dependencies and move to utils

README.md
```diff
@@ -508,12 +508,16 @@ competition that contains tags for approx 14M publications from PubMed.
 
 ## 🐍 Venv
 
-To create and setup the development environment
+To create and setup the base environment
```
nsorros, line 511: You need to move the commands around and introduce a new header development as described in the issue. The user needs to know which commands work out of the box compared to what needs additional installation

Makefile
pdan93: good catch, seems I overwrote instead of copying that one. there is an update-requirements and an update-requirements-dev now

nsorros:
Also tests should pass

nsorros:
When we get tests to pass and implement a recipe to update both requirements, let's mark the tests according to which dependendencies they need and run the tests that need the dev dependencies by installing the dev virtualenv while running the other with installing the default ones to ensure that the repo will work with the light deps.

AndreiXYZ:
Build is now passing. 
Tests are working, including locally via running `pytest`.
@nsorros Please review the changes I made when you have time

nsorros:
Are those dependencies needed in `unpinned_requirements` for `predict` and `download` to work? If not move to dev.
```
matplotlib
gensim==4.0.0
scispacy
scikit-multilearn
streamlit
seaborn
```

Makefile
```diff
@@ -79,6 +75,42 @@ update-requirements: ## Updates requirement
...
+	@mkdir -p $(VIRTUALENV)
+	virtualenv --python $(PYTHON) $(VIRTUALENV)
+	$(PIP) install --upgrade pip
+	$(PIP) install pytest pytest-cov tox
+	$(PIP) install -r dev_requirements.txt
+	$(PIP) install --no-deps -e .[dev]
+	$(PIP) install https://s3-us-west-2.amazonaws.com/ai2-s2-scispacy/releases/v0.4.0/en_core_sci_sm-0.4.0.tar.gz
+	$(VIRTUALENV)/bin/pre-commit install --hook-type pre-push --hook-type post-checkout --hook-type pre-commit
+
+update-requirements: VIRTUALENV := /tmp/update-requirements-venv/
```
nsorros, line 90: this is a duplicate and it creates a problem when run `make update-requirements`
```
Makefile:92: warning: overriding commands for target `update-requirements'
Makefile:68: warning: ignoring old commands for target `update-requirements'
```

so keep only one of them

dev_requirements.txt
```diff
@@ -0,0 +1,262 @@
...
+multiprocess==0.70.14
+murmurhash==1.0.9
+nanotime==0.5.2
+nervaluate==0.1.8
+networkx==3.1
+nmslib==2.1.1
+nodeenv==1.8.0
+numba==0.56.4
+numpy==1.19.5
+nvidia-cublas-cu11==11.10.3.66
```
nsorros, line 125: these libraries do not work in my computer, i had this problem also with daniel. did you update the requirements?

nsorros:
also create two sets of the tests that you run independently using pytest mark. for the light installation run only the predict / download tests and for all other tests run everything. for the light only install the light version

nsorros (changes requested):
- Fix makefile to remove unneeded deps
- Mark tests separately for each installation and run seprate tests installing the light version and everything

nsorros:
Also I do not think `download` or `predict` works. Can you provide a sample that works?

In `download` case there is a wrong version in the package but even when using `version = "0.2.4"` this only downloads xlinear. If anything we should download bertmesh although we can discuss whether that is needed at this point.

In `predict` case it requires a path to a label binarizer which is not needed for bert mesh. I run `grants_tagger predict malaria Wellcome/WellcomeBertMesh models/xlinear/label_binarizer-2022.12.0.pkl` so i used `xlinear` label binarizer but this did not work either.

For download I suggest

- we remove the download model since it is not needed anymore since the production model is a transformer and coming from the hub. we make download to default to epmc so we remove the subcommand and update the documentation.
- we make predict to work only with wellcome bert mesh which is our production model so we remove the need for label binarizer and we update the documentation

Makefile
AndreiXYZ: removed duplicate

nsorros:
We should also use this opportunity to simplify the default requirements. Here is the list I used locally that mostly worked
```
pandas
xlrd
scikit-learn
numpy
transformers
scipy
wasabi
typer
tqdm
requests
openpyxl
torch
```

this is for unpinned requirements.

nsorros:
After all that is done, also check which packages are taking most of the space in the virtualenv, say top 5? For me at this point these are
```
562912  venv/lib/python3.8/site-packages//torch
208728  venv/lib/python3.8/site-packages//scipy
117504  venv/lib/python3.8/site-packages//pandas
117160  venv/lib/python3.8/site-packages//transformers
111576  venv/lib/python3.8/site-packages//numpy
```

the most heavy is torch which for me is 500GB but for some linux variants gets into the GBs. It would be good to force a cpu installation of torch for inference in order to make this really light 🪶

AndreiXYZ:
> also create two sets of the tests that you run independently using pytest mark. for the light installation run only the predict / download tests and for all other tests run everything. for the light only install the light version

The `predict` test I decided to skip for now. It was a bit too much hassle to modify it to the new model

dev_requirements.txt
AndreiXYZ: The ones you highlighted seem to have been added by Daniel in [this commit](https://github.com/wellcometrust/grants_tagger/pull/246/commits/0b6e7c12988ec2c95b41e98ab77d7057f5fe9d0f)

I didn't update the dev ones. Just ran `make update-requirements` after adding the `tensorflow-addons` in the unpnined dev requirements.

It runs fine for me. What error are you getting specifically?

AndreiXYZ:
You can run tests reserved for inference time via:
`pytest -m inference_time`

nsorros (changes requested):
- You need to update the requirements files using the `make` commands
- You also need to change Travis to run tests with the light deps separately from the other tests. In the light tests you should only install the light deps

nsorros:
Also the `cuda` libraries are not removed. Add `grep -v ` in the make recipe.
```
ERROR: Could not find a version that satisfies the requirement nvidia-cublas-cu11==11.10.3.66 (from versions: 0.0.1.dev5, 0.0.1)
ERROR: No matching distribution found for nvidia-cublas-cu11==11.10.3.66
```

Makefile
```diff
@@ -76,14 +72,45 @@ update-requirements: ## Updates requirement
 	$(VIRTUALENV)/bin/pip install -r unpinned_requirements.txt
 	echo "#Created by Makefile. Do not edit." > requirements.txt
 	$(VIRTUALENV)/bin/pip freeze | grep -v pkg_resources==0.0.0 >> requirements.txt
+	grep -v nvidia-* < requirements.txt > requirements_no_nvidia.txt
```
nsorros, line 75: you want this line to be above to update the `requirements.txt`

```
$(VIRTUALENV)/bin/pip freeze | grep -v pkg_resources==0.0.0 | grep -v nvidia-* >> requirements.txt
```

Makefile
```diff
@@ -57,13 +57,9 @@ virtualenv: ## Creates virtualenv
...
 	virtualenv --python $(PYTHON) $(VIRTUALENV)
 	$(PIP) install --upgrade pip
-# As to why no identation see https://stackoverflow.com/questions/4483313/make-error-for-ifeq-syntax-error-near-unexpected-token
-ifeq ($(UNAME), Linux)
-	$(PIP) install libpecos==0.1.0
-endif
 	$(PIP) install pytest pytest-cov tox
 	$(PIP) install -r requirements.txt
-	$(PIP) install --no-deps -e .
+	$(PIP) install -e .
```
nsorros, line 62: keep the `--no-deps` and if you need extra deps you need to add in `unpinned` and run `make update-requirements`

Makefile
```diff
@@ -76,14 +72,45 @@ update-requirements: ## Updates requirement
 	$(VIRTUALENV)/bin/pip install -r unpinned_requirements.txt
 	echo "#Created by Makefile. Do not edit." > requirements.txt
 	$(VIRTUALENV)/bin/pip freeze | grep -v pkg_resources==0.0.0 >> requirements.txt
+	grep -v nvidia-* < requirements.txt > requirements_no_nvidia.txt
+	mv requirements_no_nvidia.txt requirements.txt
 #	echo "-e git://github.com/wellcometrust/WellcomeML.git@tokenizer-decode#egg=wellcomeml" >> requirements.txt
 	echo "git+https://github.com/nsorros/shap.git@dev" >> requirements.txt
```
nsorros, line 78: this is not needed for inference

Makefile
```diff
@@ -76,14 +72,45 @@ update-requirements: ## Updates requirement
...
+update-requirements-dev: VIRTUALENV := /tmp/update-requirements-venv/
+update-requirements-dev: ## Updates requirement
+	@if [ -d $(VIRTUALENV) ]; then rm -rf $(VIRTUALENV); fi
+	@mkdir -p $(VIRTUALENV)
+	virtualenv --python $(PYTHON) $(VIRTUALENV)
+	$(VIRTUALENV)/bin/pip install --upgrade pip
+	$(VIRTUALENV)/bin/pip install -r unpinned_requirements.txt -r unpinned_dev_requirements.txt
+	echo "#Created by Makefile. Do not edit." > dev_requirements.txt
+	$(VIRTUALENV)/bin/pip freeze | grep -v pkg_resources==0.0.0 >> dev_requirements.txt
+	grep -v nvidia-* < requirements.txt > requirements_no_nvidia.txt
```
nsorros, line 101: same

nsorros:
We are not at 1.4GB size for virtualenv
```
1424296	venv/lib/python3.8/site-packages/
730120	venv/lib/python3.8/site-packages/torch
92672	venv/lib/python3.8/site-packages/pydantic
87564	venv/lib/python3.8/site-packages/scipy
67852	venv/lib/python3.8/site-packages/sympy
63360	venv/lib/python3.8/site-packages/pandas
```

PR #256, opened 2023-05-29
255 get evaluation to run with new model

AndreiXYZ:
Edit the `grants_tagger evaluate model` command to use Bert Mesh.
Removed unnecessary sparse matrix support. 
Added script to convert the publications json file we received into the expected format for the evaluate function.
Change readme accordingly.


grants_tagger/evaluate_pubs.py
```diff
@@ -0,0 +1,127 @@
...
+def evaluate_pubs_cli(
+    data_path: str = typer.Argument(..., help="Path to data"),
+    threshold: float = typer.Option(0.5, help="Threshold for classification"),
+    batch_size: int = typer.Option(4, help="Batch size"),
+    device: str = typer.Option("cpu", help="Device to run on. CPU or CUDA"),
+):
+    evaluate_pubs(data_path, threshold, batch_size, device)
+
+
+if __name__ == "__main__":
```
nsorros, line 126: i thought the idea was to use `evaluate` command?

grants_tagger/evaluate_pubs.py
```diff
@@ -0,0 +1,127 @@
...
+        print("[bold red]CUDA is not available, using CPU")
+        device = "cpu"
+
+    with open(data_path, "r") as f:
+        data = json.load(f)
+
+    # Filter publications without tags
+    data = [pub for pub in data if pub["mesh_terms"]]
+
+    tokenizer = AutoTokenizer.from_pretrained("Wellcome/WellcomeBertMesh")
```
nsorros, line 28: this is hardcoded, ideally you should pass as parameter, you can name it `model_path` and make it to default to this, thats fine

grants_tagger/evaluate_pubs.py
```diff
@@ -0,0 +1,127 @@
...
+            batch_abstracts,
+            padding="max_length",
+            truncation=True,
+            max_length=512,
+            return_tensors="pt",
+        )
+
+        inputs = inputs.to(device)
+
+        outs = model(**inputs, return_labels=False)
```
nsorros, line 83: you are also not using the predict function so this is yet another duplication, we dont want to rewrite predict code for this mode multiple times.

nsorros (changes requested):
There are a couple of issues here

- this introduces an unnecessary new command that introduces duplication in code
- it further duplicates inside the command by rewriting code for predict
- it also works with one file and one model so its not generic at all

working with one model, the one in production is fine but working with one file format is  not. we should make it to work with a generic format and preprocess the file in scripts

requirements.txt
```diff
@@ -35,3 +35,4 @@ typing_extensions==4.6.1
 tzdata==2023.3
 urllib3==1.26.16
 wasabi==1.1.1
+rich==13.3.5
```
nsorros, line 38: how did that find its way here? it should be through some unpinned_requirements

scripts/convert_to_jsonl.py
```diff
@@ -0,0 +1,63 @@
+import json
+import argparse
+
+# Read json file, list of dictionaries
```
nsorros, line 4: remove those comments

nsorros (changes requested):
This is close to be ready. What remains is

- mark the evaluate test to work with the light installation and make tests pass
- update unpinned_requirements and run to update the requirements file if you need rich

nsorros (approved):
LGTM 👍

PR #260, opened 2023-05-31
add possibility to do batched inference and pre-annotate the grants

AndreiXYZ:
Add support for batch inference in the predict command.
Add possiblity to pre-annotate your grants data in the script using WellcomeBertMesh.


nsorros (approved):
Add some tests to ensure the new functionality works but other than that its good to go

PR #262, opened 2023-06-01
Add script to upload grants data to Argilla

AndreiXYZ:
Added a script to upload grants data as generated by the `create_grants_sample.py` to Argilla.
Added Argilla to unpinned and updated requirements.


nsorros (approved):
My only concern here was whether to add this into a DVC pipeline, let's keep that in mind we might do later when the annotations come back. It is just for reproducibility.

nsorros (changes requested):
Let's 

- add the for tags
- add the files in DVC

nsorros (approved):

